PR #1, opened 2022-01-24
Solution

phantom-terminal:


app/main.py
```diff
@@ -211,4 +18,5 @@ def battle(knightsConfig):
     }
 
 
-print(battle(KNIGHTS))
+if __name__ == '__main__':
+    print(battle(KNIGHTS))
```
Abnormaltype, line 22: Do you need this?

app/uitls/feature_count.py
```diff
@@ -0,0 +1,16 @@
...
+    if result["potion"] is not None:
+        if "power" in result["potion"]["effect"]:
+            result["power"] += result["potion"]["effect"]["power"]
+
+        if "protection" in result["potion"]["effect"]:
+            result["protection"] += result["potion"]["effect"]["protection"]
+
+        if "hp" in result["potion"]["effect"]:
+            result["hp"] += result["potion"]["effect"]["hp"]
+
```
Abnormaltype, line 15: These code duplications can be also fixed, using stats = ("protection", "power", "hp") and looping on stats
Ex:
`stats = ("protection", "power", "hp")`

```
if knight["potion"] is not None:

                for statistic in stats:
                    if statistic in knight["potion"]["effect"]:
                        knight[statistic] += knight["potion"]["effect"][
                            statistic
                        ]
```

app/uitls/feature_count.py
```diff
@@ -0,0 +1,16 @@
+def feature_count(knightsConfig, knight_2):
+    result = knightsConfig[knight_2]
+    result["protection"] = sum(a["protection"] for a in result["armour"])
```
Abnormaltype, line 3: a is a bad name, armour is better in this case

Abnormaltype (changes requested):

app/main.py
phantom-terminal: if __name__ == '__main__' This is the entry point. If I want to change or add data, I can run this function here (+import data) sequentially with different data.
About the print() I agree. You don't need it here.

Abnormaltype: `def battle(knightsConfig):` and pls use `snake_case` -> `knights_config` 

in main on str 6

app/uitls/feature_count.py
phantom-terminal: Wow, cool, thanks)

app/main.py
phantom-terminal: Tnx for camel Case)

Abnormaltype (approved):
Good job!

PR #2, opened 2022-01-24
Solution

Krimat:


app/main.py
```diff
@@ -86,128 +88,21 @@
 }
```
phantom-terminal, line 89: The data may be better separated from the main function

phantom-terminal (changes requested):

MasterpieceElbow (approved):

PR #3, opened 2022-01-24
Solution

Abnormaltype:


app/battle/battle.py
```diff
@@ -0,0 +1,31 @@
+class Battle:
+
+    def __init__(self, knights_dict: dict):
+        self.knights_dict = knights_dict
+
+    def versus(self, knight_1: str, knight_2: str):
+        result_dict = {}
+        if self.knights_dict.get(knight_1) and self.knights_dict.get(knight_2):
```
danylott, line 8: Why are you using .get() instead of simple [] here?

danylott: This `if` seems to be unnecessary

app/battle/battle.py
```diff
@@ -0,0 +1,31 @@
...
+            self.knights_dict[knight_1].name['hp'] -= \
+                self.knights_dict[knight_2].name['power'] \
+                - self.knights_dict[knight_1].name['protection']
+            self.knights_dict[knight_2].name['hp'] -= \
+                self.knights_dict[knight_1].name['power'] \
+                - self.knights_dict[knight_2].name['protection']
```
danylott, line 14: DRY: Don't Repeat Yourself
WET: Write Everything Twice

I believe - you choose WET here, instead of DRY?

app/knights/battle_preparation.py
```diff
@@ -0,0 +1,38 @@
...
+    def battle_preparation(self):
+        # BATTLE PREPARATIONS:
+        dict_knights = {}
+        for knight in self.knights:
+            # knight
+            knight = self.knights[knight]
+
+            # apply armour
+            knight["protection"] = 0
+            for a in knight["armour"]:
```
danylott, line 18: name `a` is bad - use `armour` instead

app/knights/battle_preparation.py
```diff
@@ -0,0 +1,38 @@
...
+            knight["power"] += knight["weapon"]["power"]
+
+            # apply potion if exist
+            if knight["potion"] is not None:
+                if "power" in knight["potion"]["effect"]:
+                    knight["power"] += knight["potion"]["effect"]["power"]
+
+                if "protection" in knight["potion"]["effect"]:
+                    knight["protection"] +=\
+                        knight["potion"]["effect"]["protection"]
```
danylott, line 31: Better:
```python
knight["protection"] += knight["potion"]["effect"][
    "protection"
]
```

app/knights/battle_preparation.py
```diff
@@ -0,0 +1,38 @@
...
+
+                if "hp" in knight["potion"]["effect"]:
+                    knight["hp"] += knight["potion"]["effect"]["hp"]
```
danylott, line 34: These code duplications can be also fixed, using `stats = ("protection", "power", "hp")` and looping on `stats`

app/knights/battle_preparation.py
```diff
@@ -0,0 +1,38 @@
...
+
+            knight_object = Knight(knight, knight["power"], knight["hp"])
```
danylott, line 36: A bit strange, that you creating Knight Object after all calculations.
Do not really understand the purpose of it here - but ok

app/main.py
```diff
@@ -1,214 +1,27 @@
...
-                "power": +5,
-            }
-        }
-    }
-}
+from app.battle.battle import Battle
+from app.knights.battle_preparation import Preparation
 
 
 def battle(knightsConfig):
```
danylott, line 5: `camelCase` in python is a bad practice - use `snake_case`

danylott (changes requested):

app/battle/battle.py
Abnormaltype: where DRY? I’m counting for first and second HP

app/knights/battle_preparation.py
Abnormaltype: oh,  sorry, my mistacke

app/knights/battle_preparation.py
Abnormaltype: okay, thx, flake8)

app/main.py
Abnormaltype: okay

app/knights/battle_preparation.py
Abnormaltype: okay, thx

app/battle/battle.py
danylott: Hm, okay, it can be a method of Knight - to beat different Knight, but I agree with - this implementation is also ok

danylott (approved):

PR #5, opened 2022-01-24
Solution

Nazar-kob:


app/battle/battle.py
```diff
@@ -0,0 +1,19 @@
+class Battle:
+
+    def __init__(self, knight_1, knight_2):
+        self.knight_1 = knight_1
+        self.knight_2 = knight_2
+
+    def start_battle(self):
+        while True:
```
Krimat, line 8: In battle each knight does one hit each other, not till someone dies.

app/my_main.py
```diff
@@ -0,0 +1,50 @@
+
+from app.knights.knight import Knight
+from app.armours.armour import Armour
+from app.weapons.weapon import Weapon
+from app.weapons.potions.potion import Potion
+from app.battle.battle import Battle
+
+
+lancelot = Knight('lancelot', 35, 100)
```
Krimat, line 9: Lots of magic const, Use KNIGHTS dictionary from main.py

app/my_main.py
```diff
@@ -0,0 +1,50 @@
+
+from app.knights.knight import Knight
```
Krimat, line 2: Rewrite this file to main.py to see if it passes tests

Krimat (changes requested):

app/battle/battle.py
Nazar-kob: Thanks for the comment, but in my case they are fighting to the end and only one will win

app/my_main.py
Nazar-kob: Thanks for the comment, but they are needed for logical implementation!

app/my_main.py
Nazar-kob: Thanks for the comment, but the test will not pass because of change the code

MasterpieceElbow (approved):
Nice job!

PR #8, opened 2022-01-25
Solution

TPolina:


app/knight.py
```diff
@@ -0,0 +1,29 @@
...
+        return self
+
+    def apply_potion_if_any(self):
+        if self.potion is not None:
+            if "power" in self.potion["effect"]:
+                self.power += self.potion["effect"]["power"]
+            if "hp" in self.potion["effect"]:
+                self.hp += self.potion["effect"]["hp"]
+            if "protection" in self.potion["effect"]:
+                self.protection += self.potion["effect"]["protection"]
```
Abnormaltype, line 29: These code duplications can be also fixed, using stats = ("protection", "power", "hp") and looping on stats

app/main.py
```diff
@@ -1,214 +1,31 @@
...
-    # lancelot
-    lancelot = knightsConfig["lancelot"]
+    if rival1.hp <= 0:
+        rival1.hp = 0
 
-    # apply armour
-    lancelot["protection"] = 0
-    for a in lancelot["armour"]:
-        lancelot["protection"] += a["protection"]
+    if rival2.hp <= 0:
```
Abnormaltype, line 11: It's not a mistake but I think it better to create a function check_hp()

Abnormaltype (changes requested):
Nice work!

app/main.py
```diff
@@ -1,214 +1,31 @@
...
-        if "power" in lancelot["potion"]["effect"]:
-            lancelot["power"] += lancelot["potion"]["effect"]["power"]
 
-        if "protection" in lancelot["potion"]["effect"]:
-            lancelot["protection"] += lancelot["potion"]["effect"]["protection"]
+def battle(knights):
+    knights = {prsn["name"]: Knight(
+        prsn["name"], prsn["power"], prsn["hp"],
+        prsn["armour"], prsn["weapon"], prsn["potion"])
+        for prsn in knights.values()}
```
Abnormaltype, line 21:
```
def battle(knights):
    knights = {
        prsn["name"]: Knight(
            prsn["name"], prsn["power"], prsn["hp"],
            prsn["armour"], prsn["weapon"], prsn["potion"]
        )
        for prsn in knights.values()
    }
```
It's more readable

Abnormaltype (changes requested):

app/knight.py
TPolina: don't really understand how to use this, can you give a clue?

Following code will not work because of self.stat:
_for stat in ("power", "hp", "protection"):
    if stat in self.potion["effect"]:
        self.stat += self.potion["effect"][stat]_

Abnormaltype:
```
stats = ("protection", "power", "hp")

if knight["potion"] is not None:

                for statistic in stats:
                    if statistic in knight["potion"]["effect"]:
                        knight[statistic] += knight["potion"]["effect"][
                            statistic
                        ]
```

Abnormaltype: Hmm..
I didn’t notice self.
We’ll probably have to leave it at that.

Abnormaltype (changes requested):
Good job!
`setattr(self, stat, getattr(self, stat) + self.potion["effect"][stat])` very well solution!

Abnormaltype (approved):

app/main.py
```diff
@@ -1,214 +1,35 @@
...
-        lancelot["name"]: lancelot["hp"],
-        arthur["name"]: arthur["hp"],
-        mordred["name"]: mordred["hp"],
-        red_knight["name"]: red_knight["hp"],
+def battle(knights):
+    knights = {
+        prsn["name"]: Knight(
+            prsn["name"], prsn["power"], prsn["hp"],
+            prsn["armour"], prsn["weapon"], prsn["potion"])
+        for prsn in knights.values()
```
MasterpieceElbow, line 24: `prsn` is very bad loop variable name. Pick a more relevant name.

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #16, opened 2022-05-19
Develop

DmytroSvirsa:


app/knights/preparation.py
```diff
@@ -0,0 +1,19 @@
+class Knight:
+    def __init__(self, knight_and_amo: dict):
+        self.name = knight_and_amo["name"]
+        self.power = knight_and_amo["power"]
+        self.hp = knight_and_amo["hp"]
+        self.dr = 0
```
Magus6784, line 6: I think dr = defense rate but I've been playing video games all my life. Bad name for protection

app/knights/preparation.py
```diff
@@ -0,0 +1,19 @@
+class Knight:
+    def __init__(self, knight_and_amo: dict):
+        self.name = knight_and_amo["name"]
+        self.power = knight_and_amo["power"]
+        self.hp = knight_and_amo["hp"]
+        self.dr = 0
+
+        for a in knight_and_amo["armour"]:
+            self.dr += a["protection"]
```
Magus6784, line 9: I think its okey for one use
But rename "a" to "armour" good idea (or not)

Magus6784, line 9: I think its okey for one use
But rename "a" to "armour" good idea (or not)

Magus6784 (changes requested):
I dont know about use so many thing in __init__. Maybe create another func or method for this, but up to you

app/knights/preparation.py
DmytroSvirsa: it's not from computer games it's from tabletop RPG games GURPS, but ep, you are right

Magus6784 (approved):

MasterpieceElbow (approved):

PR #17, opened 2022-05-19
'Solution'

Magus6784:


app/fight.py
```diff
@@ -0,0 +1,17 @@
+def fight(fighters: dict):
+    combat(fighters["lancelot"], fighters["mordred"])
```
KatiukhaO, line 2: inside this function, you use keys "lancelot", "mordred"....
but, this names include to knights, 
if look on module fight separtly, and dont now about other modules, it may become unclear what kind of data it is and what the general logic of the project is
so maybe better functions combat and chek_death to create methods of class RegistrationKnight

KatiukhaO (changes requested):

app/fight.py
Magus6784: I dont think so
Maybe king say "Hey, i want see lion at my tournament" and at this case we can use
combat(fighters["Simba"], fighters["Scar"])
And this can work if class Lion have .hp, .power. protection

Magus6784: Yea, this fight bound with another code, but if we have new type of fighters, we can recreate function without doublecode

KatiukhaO (approved):

app/fight.py
KatiukhaO: ok, I meant a little different
i like your code)
i approve, because, it's will be holy war, two knights of Mate Academy)

MasterpieceElbow (approved):

PR #21, opened 2022-05-20
Made 1 module with all methods needed for battle. Now it is possible …

Hectorovich:
…to compete any number of knights

### DIFF
--- a/app/battle.py
+++ b/app/battle.py
@@ -1,0 +1,48 @@
+class Battle:
+    def __init__(self, profile):
+        self.profile = profile
+
+    def armour_preparation(self):
+        for stat in self.profile.values():
+            stat["protection"] = 0
+            for a in stat["armour"]:
+                stat["protection"] += a["protection"]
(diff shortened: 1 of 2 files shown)


app/battle.py
```diff
@@ -0,0 +1,48 @@
+class Battle:
+    def __init__(self, profile):
+        self.profile = profile
+
+    def armour_preparation(self):
+        for stat in self.profile.values():
+            stat["protection"] = 0
+            for a in stat["armour"]:
+                stat["protection"] += a["protection"]
+        return self.profile
```
Chardona, line 10: I think it's better to pass required elements to the function and return them, than work with entire set of object atributes.

Chardona (approved):
I think it would've been better to separate knights into different class objects and add methods to operate with them. In your case you work with only 1 class, which contain knights as atributes. It works for tests but practically it's hard to work with.

TPolina (approved):

PR #23, opened 2022-05-22
Monty Python Battle

Chardona:


app/main.py
```diff
@@ -1,213 +1,53 @@
...
+                        cfg["red_knight"]["hp"],
+                        cfg["red_knight"]["armour"],
+                        cfg["red_knight"]["weapon"],
+                        cfg["red_knight"]["potion"])
+
+    # preparation
+    lancelot.equip().drink_potion()
+    arthur.equip().drink_potion()
+    mordred.equip().drink_potion()
+    red_knight.equip().drink_potion()
```
TPolina, line 40: can you optimize this with the help of cycles or something else, please

TPolina (changes requested):

app/main.py
Chardona: well, that was my view of knight's show  presentation, but i'll optimize that.

MasterpieceElbow (approved):

PR #26, opened 2022-05-26
Solution

MykolaIF:


app/knight_class.py
```diff
@@ -0,0 +1,24 @@
+class Knight:
+    def __init__(self, knight):
+        self.name = knight["name"]
+        self.protection = 0
+        self.power = knight["power"]
+        self.hp = knight["hp"]
+
+        # apply armour
+        for a in knight["armour"]:
```
IvanRamyk, line 9: name `a` is bad - use `armour` instead.

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #36, opened 2022-06-27
Solution

anatomst:


app/characters/battle_preparation.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def battle_preparation(knights):
+    knights_dict = {}
+
+    for character in knights:
+        hp = knights[character]["hp"]
+
+        protection = 0
+        for a in knights[character]["armour"]:
```
Abnormaltype, line 11: name `a` is bad - use `armour` instead

app/characters/battle_preparation.py
```diff
@@ -0,0 +1,31 @@
...
+        if knights[character]["potion"] is not None:
+            if "power" in knights[character]["potion"]["effect"]:
+                power += knights[character]["potion"]["effect"]["power"]
+
+            if "protection" in knights[character]["potion"]["effect"]:
+                protection += \
+                    knights[character]["potion"]["effect"]["protection"]
+
+            if "hp" in knights[character]["potion"]["effect"]:
+                hp += knights[character]["potion"]["effect"]["hp"]
```
Abnormaltype, line 26: These code duplications can be also fixed, using `stats = ("protection", "power", "hp")` and looping on `stats`

Abnormaltype (changes requested):

app/characters/battle_preparation.py
anatomst: how*

Abnormaltype: write me in slack, your versions

Abnormaltype (approved):
well done!

PR #40, opened 2022-07-08
solution

tarkaiev:


app/main.py
```diff
@@ -1,3 +1,5 @@
+from app.fighters.knight import Knight
+
 KNIGHTS = {
```
Abnormaltype, line 3: move this variable to the new file

Abnormaltype (changes requested):

app/main.py
tarkaiev: Moved to new file.

Abnormaltype (approved):
well done

PR #41, opened 2022-07-08
Solution

InnaKushnir:


app/knights/apply_armour.py
```diff
@@ -0,0 +1,8 @@
+class Armour:
+
+    @staticmethod
+    def app_armour(armour: dict):
+        protection = 0
+        for a in armour:
```
Abnormaltype, line 6: `a` is an uninformative name for variable

app/knights/apply_potion.py
```diff
@@ -0,0 +1,13 @@
...
+    @staticmethod
+    def app_potion(potion: dict, power, hp, protection):
+        if "power" in potion["effect"]:
+            power += potion["effect"]["power"]
+
+        if "protection" in potion["effect"]:
+            protection += potion["effect"]["protection"]
+
+        if "hp" in potion["effect"]:
+            hp += potion["effect"]["hp"]
```
Abnormaltype, line 12: These code duplications can be fixed, using `stats = ("protection", "power", "hp")` and looping on `stats`

app/main.py
```diff
@@ -86,129 +89,47 @@
...
 
-        if "hp" in arthur["potion"]["effect"]:
-            arthur["hp"] += arthur["potion"]["effect"]["hp"]
+    # Preaparations
+    # list of lists [name, power, hp, protection]
+    Knights = []
 
-    # mordred
-    mordred = knightsConfig["mordred"]
+    for k, v in KnightsConfig.items():
```
Abnormaltype, line 98: what are `k` and `v`?

app/main.py
```diff
@@ -86,129 +89,47 @@
...
+    result = []
+    for i in range(2):
+        result.append(Knights[i][2] - Knights[i + 2][1] + Knights[i][3])
+        print(result)
+        result.append(Knights[i + 2][2] - Knights[i][1] + Knights[i + 2][3])
 
-    # check if someone fell in battle
-    if lancelot["hp"] <= 0:
-        lancelot["hp"] = 0
+    for j in range(4):
```
Abnormaltype, line 126: uninformative loop, why 4?

Abnormaltype (changes requested):

app/main.py
InnaKushnir: 4, becouse in this case 4 knights. I can change this value

app/main.py
InnaKushnir: k - key, v- value in dictionary

Abnormaltype: please, write this in code)

app/main.py
Abnormaltype: create more informative loop

Abnormaltype (approved):
well done

app/knights/apply_armour.py
InnaKushnir: return self.hidden is True if self.hidden is False else self.hidden is False
When I write in one line  - not pass all tests

InnaKushnir: That comment shouldn't have been there

app/main.py
```diff
@@ -86,129 +89,47 @@
...
-            arthur["hp"] += arthur["potion"]["effect"]["hp"]
+    # Preaparations
+    # list of lists [name, power, hp, protection]
+    Knights = []
 
-    # mordred
-    mordred = knightsConfig["mordred"]
+    for key, value in KnightsConfig.items():
+        # list [name, power, hp, protection)
+        Knight = []
```
Abnormaltype, line 100: use `knight`  instead of `Knight`, because I mean `Knight`, it's a class

Abnormaltype (changes requested):

app/main.py
Abnormaltype: then it would be better `knight_list`

app/main.py
```diff
@@ -86,129 +89,47 @@
...
-
-        if "protection" in arthur["potion"]["effect"]:
-            arthur["protection"] += arthur["potion"]["effect"]["protection"]
+def battle(KnightsConfig):
 
-        if "hp" in arthur["potion"]["effect"]:
-            arthur["hp"] += arthur["potion"]["effect"]["hp"]
+    # Preaparations
+    # list of lists [name, power, hp, protection]
+    Knights = []
```
Abnormaltype, line 96: what about this?

Abnormaltype (changes requested):

app/main.py
InnaKushnir: This is a list , where knight_list lists of information about knights are stored

Abnormaltype (changes requested):
all tests should are pass

Abnormaltype (approved):
well done

PR #48, opened 2022-07-25
solution

hikehikehike:


app/main.py
```diff
@@ -1,213 +1,64 @@
...
+        self.power = power
+        self.hp = hp
+        self.protection = protection
 
-    # apply potion if exist
-    if arthur["potion"] is not None:
-        if "power" in arthur["potion"]["effect"]:
-            arthur["power"] += arthur["potion"]["effect"]["power"]
+    def new_stat(self, name):
+        for a in name["armour"]:
```
Abnormaltype, line 12: use `armour` instead of `a`

app/main.py
```diff
@@ -1,213 +1,64 @@
...
-    mordred = knightsConfig["mordred"]
+            if "protection" in name["potion"]["effect"]:
+                self.protection += name["potion"]["effect"]["protection"]
 
-    # apply armour
-    mordred["protection"] = 0
-    for a in mordred["armour"]:
-        mordred["protection"] += a["protection"]
+            if "hp" in name["potion"]["effect"]:
+                self.hp += name["potion"]["effect"]["hp"]
```
Abnormaltype, line 27: These code duplications can be also fixed, using `stats = ("protection", "power", "hp")` and looping on `stats`

Abnormaltype (changes requested):

app/main.py
hikehikehike: done

Abnormaltype (changes requested):
all tests should are pass

app/main.py
hikehikehike: I don't understand how to do it. there is only one idea, but it does not work)

Abnormaltype: Try to use this:
`setattr(self, stat, getattr(self, stat) + self.potion["effect"][stat])`

Abnormaltype (approved):
Well done!

PR #51, opened 2022-07-30
'Solution'

Viktor-Beniukh:


app/fight/knightConfig.py
```diff
@@ -0,0 +1,32 @@
...
+            self.protection += armour["protection"]
+        if self.potion is not None:
+            stats = ("protection", "power", "hp")
+            for stat in stats:
+                if stat in self.potion["effect"]:
+                    setattr(self,
+                            stat,
+                            getattr(self, stat) + self.potion["effect"][stat])
+
+    def order_def_winner(self, other):
```
IvanRamyk, line 26: What does `def` mean?

IvanRamyk (changes requested):

app/fight/knightConfig.py
Viktor-Beniukh: Sorry. It's a small mistake. I've already corrected.

app/fight/knightConfig.py
```diff
@@ -0,0 +1,32 @@
...
+            self.protection += armour["protection"]
+        if self.potion is not None:
+            stats = ("protection", "power", "hp")
+            for stat in stats:
+                if stat in self.potion["effect"]:
+                    setattr(self,
+                            stat,
+                            getattr(self, stat) + self.potion["effect"][stat])
+
+    def order_det_winner(self, other):
```
IvanRamyk, line 26: And what is det?)
It seems that it's a reduction. Could you write the whole word?

IvanRamyk (changes requested):
Everything else is OK, Good job!

app/fight/knightConfig.py
Viktor-Beniukh: Hi. 'det' == 'determination'.

IvanRamyk (approved):
Well done!

PR #64, opened 2022-08-27
'Solution'

davidkrivko:


app/main.py
```diff
@@ -1,214 +1,4 @@
...
-        "potion": {
-            "name": "Blessing",
-            "effect": {
-                "hp": +10,
-                "power": +5,
-            }
-        }
-    }
-}
+import battle
```
IvanRamyk, line 1:
```suggestion
import app.battle
```

app/battle.py
```diff
@@ -0,0 +1,37 @@
+from knights import Knight
```
IvanRamyk, line 1:
```suggestion
from app.knights import Knight

```

IvanRamyk (approved):
Please, fix the imports to pass tests

PR #67, opened 2022-09-01
'Solution'

MilManIT:
Помогите пожалуйста найти ошибку.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -3,7 +3,7 @@ import app.battle_preparation as b_p
 import app.battle as bat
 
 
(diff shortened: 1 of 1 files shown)


IvanRamyk (changes requested):
Usually, we use lowercase names for modules
Something wrong with imports. Tests weren't passed

DmytroSvirsa (changes requested):
nothing canged Usually, we use lowercase names for modules
Something wrong with imports. Tests weren't passed

DmytroSvirsa (approved):

PR #68, opened 2022-09-02
'Solution'

rakamakaphone:


preparation/stats.py
```diff
@@ -0,0 +1,24 @@
+# from preparation.knights import KNIGHTS
```
IvanRamyk, line 1: Please, remove the redundant comment

IvanRamyk (approved):
Well done!

PR #71, opened 2022-09-05
Solution

Serhii-Kovalenko44:


DmytroSvirsa (changes requested):
let's remove comments, now we have a clear structure, and they seem to be unneeded

DmytroSvirsa (approved):

PR #76, opened 2022-09-08
Solution

vlad-milkov:


app/result.py
```diff
@@ -0,0 +1,8 @@
+from app.knight import Knight
+
+
+class Result:
+    @classmethod
+    def return_result(cls, knights: list[Knight]):
+        result = {knight.name: knight.hp for knight in knights}
+        return result
```
DmytroSvirsa, line 8: make it as a function, if you want but there is no need to use class and classmethod

app/battle.py
```diff
@@ -0,0 +1,15 @@
+from app.knight import Knight
+
+
+class Battle:
+    @classmethod
+    def two_knights_battle(cls, knights: list[Knight]):
+        for i in range(len(knights)):
+            knights[i].hp -= \
+                knights[(i + 1) % len(knights)].power - knights[i].protection
+            Battle.fix_hp_if_fell(knights[i])
```
DmytroSvirsa, line 10: You need class method when you want to operate with class, but you don't use it, so it can be just function

DmytroSvirsa (changes requested):

DmytroSvirsa (approved):

PR #79, opened 2022-09-14
'Solution'

YuliyaGerasymniuk:


app/knight/knight.py
```diff
@@ -0,0 +1,67 @@
...
+            else:
+                potion_hp = 0
+            if "power" in potion_dict:
+                potion_power = potion_dict["power"]
+            else:
+                potion_power = 0
+            if "protection" in potion_dict:
+                potion_protection = potion_dict["protection"]
+            else:
+                potion_protection = 0
```
Abnormaltype, line 52: DRY, create a loop for this action

Abnormaltype (changes requested):

Abnormaltype (changes requested):
all tests should pass. Wrote in the slack

app/knight/battles.py
```diff
@@ -0,0 +1,24 @@
...
+    if lancelot["hp"] <= 0:
+        lancelot["hp"] = 0
+
+    if mordred["hp"] <= 0:
+        mordred["hp"] = 0
+
+    return lancelot, mordred
+
+
+def second_battle(arthur, red_knight):
```
IvanRamyk, line 14: It seems that this function is a copy of the `first_battle` with different names for arguments. It's not ok

IvanRamyk: You can create one function and pass different knights there twice

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

IvanRamyk (approved):
Well done!

PR #82, opened 2022-09-19
Solution

yessloss:


app/main.py
```diff
@@ -86,129 +86,35 @@
...
+        for armour_knight in knight["armour"]:
+            knight["protection"] += armour_knight["protection"]
+    for knight_power in list_knights:
+        knight_power["power"] += knight_power["weapon"]["power"]
+    for knight_magic in list_knights:
+        if knight_magic["potion"] is not None:
+            for value in list_power:
+                if value in knight_magic["potion"]["effect"]:
+                    knight_magic[value] += \
+                        knight_magic["potion"]["effect"][value]
```
IvanRamyk, line 107: Probably it should be a separate method/function to provide this logic

IvanRamyk (changes requested):

app/main.py
yessloss: why

IvanRamyk (approved):

PR #83, opened 2022-09-19
Preliminary solution

StasOkhrym:


app/main.py
```diff
@@ -1,7 +1,7 @@
 from app.knights import Knight
 
 
-def battle(knights):
+def battle(knights) -> dict:
```
wQuelS, line 4: add type annotation to arg:)

Everything else looks OK, imo

wQuelS (changes requested):

app/main.py
StasOkhrym: yes, sir!

AriohBelskij (approved):
its work. but i think u can do this better, u can make ur code cleaner, for example dont write hp.check() method in Knight class, and just add 1 if in attack_enemy. But its my opinion, mentor will tell u what is True

Y-Havryliv (approved):
GJ!

PR #100, opened 2022-09-23
Solution "Knights-fighting" was added

Ihor-Horiainov:


app/main.py
```diff
@@ -86,128 +88,54 @@
...
-
-    # apply armour
-    red_knight["protection"] = 0
-    for a in red_knight["armour"]:
-        red_knight["protection"] += a["protection"]
-
-    # apply weapon
-    red_knight["power"] += red_knight["weapon"]["power"]
+def knight_dict_create(data: dict) -> dict:
+    answer = dict()
```
IvanRamyk, line 92: `answer` isn't a good name :( 
It doesn't show what is stored in it

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #104, opened 2022-09-26
Solution

HrytVlad:


app/battle_begin/result.py
```diff
@@ -0,0 +1,7 @@
+class Result:
+    result = {}
+
+    @staticmethod
+    def show_result(knights):
+        Result.result = ({knight.name: knight.hp for knight in knights})
+        return Result
```
DmytroSvirsa, line 7: that is too much (simple is better than complex)

app/before_battle/basic_stats.py
```diff
@@ -0,0 +1,19 @@
+class NewKnight:
```
DmytroSvirsa, line 1: why New? that is just Knight or LegendaryKnight

PR #112, opened 2022-10-09
Solution

VadymShkarbul:


app/equipment.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def drink_potion(self) -> list:
+        potion_effect = [0, 0, 0]
+
+        if self.potion is not None:
+            effect = self.potion["effect"]
+        else:
+            return potion_effect
+
+        if effect.get("power"):
```
pazyn404, line 29: You can use get with default value, without if condition

app/equipment.py
```diff
@@ -0,0 +1,36 @@
...
+            potion_effect[0] += effect["power"]
+        if effect.get("hp"):
```
pazyn404, line 31: Here the same

app/equipment.py
```diff
@@ -0,0 +1,36 @@
...
+            potion_effect[1] = effect["hp"]
+        if effect.get("protection"):
```
pazyn404, line 33: Here the same

app/main.py
```diff
@@ -1,214 +1,40 @@
...
-    # apply armour
-    red_knight["protection"] = 0
-    for a in red_knight["armour"]:
-        red_knight["protection"] += a["protection"]
+    arthur.hp -= red_knight.power - arthur.protection
+    red_knight.hp -= arthur.power - red_knight.protection
 
-    # apply weapon
-    red_knight["power"] += red_knight["weapon"]["power"]
+    arthur.hp = health_check(arthur.hp)
```
pazyn404, line 32: You can use, just max function with 0 and current hp instead of health_check.

pazyn404 (changes requested):

app/equipment.py
pazyn404: You can tou get with other 0 default value, but it's up to you. I approved second commit.

app/equipment.py
```diff
@@ -0,0 +1,36 @@
+class Armor:
+    def __init__(self, armour: list) -> None:
+        self.armour = armour
+
+    def apply_armor(self) -> int:
+        return sum([part["protection"] for part in self.armour])
```
Nattalli, line 6: you can remove square brackets here, it would be more readable

Nattalli (changes requested):

app/equipment.py
VadymShkarbul: Oh, yes! Thank you!

PR #122, opened 2022-10-27
Develop

Soobig666:


app/preparation.py
```diff
@@ -0,0 +1,66 @@
...
+
+    # apply weapon
+    red_knight["power"] += red_knight["weapon"]["power"]
+
+    # apply potion if exist
+    if red_knight["potion"] is not None:
+        for effect in red_knight["potion"]["effect"]:
+            red_knight[effect] += red_knight["potion"]["effect"][effect]
+
+    return knights_config
```
Abnormaltype, line 66: DRY, you can create a function for knight actions

Abnormaltype (changes requested):

app/preparation.py
```diff
@@ -0,0 +1,14 @@
+def get_ready(knights_config: dict) -> dict:
+    for person in knights_config.keys():
+        king = knights_config[person]
```
IvanRamyk, line 3: Why king?

app/preparation.py
```diff
@@ -0,0 +1,14 @@
+def get_ready(knights_config: dict) -> dict:
+    for person in knights_config.keys():
```
IvanRamyk, line 2:
```suggestion
    for person in knights_config:
```

app/preparation.py
```diff
@@ -0,0 +1,14 @@
+def get_ready(knights_config: dict) -> dict:
```
IvanRamyk, line 1: I think it isn't the best function name :( 
Maybe smth like `set_power_and_protection` is more descriptive

app/main.py
```diff
@@ -1,213 +1,29 @@
...
-    if mordred["hp"] <= 0:
-        mordred["hp"] = 0
 
-    # 2 Arthur vs Red Knight:
-    arthur["hp"] -= red_knight["power"] - arthur["protection"]
-    red_knight["hp"] -= arthur["power"] - red_knight["protection"]
+def battle(main_dict_knights: dict) -> dict:
+    preparation = get_ready(main_dict_knights)
+    count = 0
+    heroes = [person for person in preparation]
```
IvanRamyk, line 9: it seems that `heroes` is just a duplication

IvanRamyk (changes requested):

app/main.py
IvanRamyk: It seems that heroes is equal to preparation and you can just use `preparation` list

IvanRamyk: just check values of these variables in debugger :)

IvanRamyk (changes requested):

app/preparation.py
YuriiKindrat: u should change king to knight everywhere, cause in future u should write accurate names of variables

YuriiKindrat (changes requested):

Abnormaltype (approved):

PR #124, opened 2022-10-30
knights-fighting

Mazurinka29:


app/main.py
```diff
@@ -86,129 +94,48 @@
 }
 
 
-def battle(knightsConfig):
+def battle(knights_config: dict) -> dict:
     # BATTLE PREPARATIONS:
+    lancelot = knights_config["lancelot"]
```
IvanRamyk, line 99: Let's use a loop to avoid duplication

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #127, opened 2022-11-03
Solutions

Foksanv:


app/knight/warior.py
```diff
@@ -0,0 +1,52 @@
...
+
+from app.knight_abilities.potion import Potion
+from app.knight_abilities.weapon import Weapon
+from app.knight_abilities.armor import Armour
+
+
+class Knight:
+    def __init__(self, name: str, dic_info: dict) -> None:
+        self.dic_info = dic_info
+        self.knight_info = self.dic_info[name]
```
ZhAlexR, line 11: I could not claim that the following statement  is 100 %: it seems to me that the logic of looking up a specific knight (row 11) in the dictionary should be moved outside the class.

ZhAlexR (approved):

app/knight/warior.py
YuriiKindrat: no need to save all dict info in every instance of class Knight, u can just pass the info about several knight

app/main.py
```diff
@@ -1,214 +1,21 @@
...
+def battle(config: dict) -> dict:
+    result = {}
+    lancelot = Knight("lancelot", config)
+    mordred = Knight("mordred", config)
+    artur = Knight("arthur", config)
+    red_knight = Knight("red_knight", config)
+    pairs = [[lancelot, mordred], [artur, red_knight]]
+    for pair in pairs:
+        pair[0].prepare_for_tournament()
+        pair[1].prepare_for_tournament()
```
YuriiKindrat, line 16: u can call this method inside the init method

app/main.py
```diff
@@ -1,214 +1,21 @@
...
+from app.knight.info import KNIGHTS
+from app.knight.warior import Knight
+
+
+def battle(config: dict) -> dict:
+    result = {}
+    lancelot = Knight("lancelot", config)
+    mordred = Knight("mordred", config)
+    artur = Knight("arthur", config)
+    red_knight = Knight("red_knight", config)
```
YuriiKindrat, line 12: u can use for loop for this, also no need to write their names they already saved in dict

YuriiKindrat (changes requested):

PR #144, opened 2022-12-01
Develop

roma-mechanic:


app/equipment/armour.py
```diff
@@ -0,0 +1,11 @@
+from app.keys import KeysArmour
+
+
+class Armour:
+    def __init__(self, list_armour: list) -> None:
+        protect = []
```
IvanRamyk, line 6: why create a list? 
if you just want to calculate a sum use an integer variable

IvanRamyk (approved):

PR #145, opened 2022-12-02
Solution

Isles22:


app/preparation/preparation.py
```diff
@@ -0,0 +1,19 @@
+from __future__ import annotations
+from app.class_knight.knight import Knight
+
+
+def preparation(knights_config: dict) -> Knight:
```
OleksandrPozden, line 5: incorrect annotation. we return somethind Knight.knights but not a Knight

app/main.py
```diff
@@ -86,129 +90,20 @@
...
+    red_knight = preparation(knights_config)["Red Knight"]
+    lancelot.hp -= mordred.power - lancelot.protection
+    mordred.hp -= lancelot.power - mordred.protection
+    artur.hp -= red_knight.power - artur.protection
+    red_knight.hp -= artur.power - red_knight.protection
+
+    return {lancelot.name: positive(lancelot.hp),
+            artur.name: positive(artur.hp),
+            mordred.name: positive(mordred.hp),
+            red_knight.name: positive(red_knight.hp)}
```
OleksandrPozden, line 106: instead of positive, you can use ```abs``` as default function

app/main.py
```diff
@@ -86,129 +90,20 @@
...
-        lancelot["name"]: lancelot["hp"],
-        arthur["name"]: arthur["hp"],
-        mordred["name"]: mordred["hp"],
-        red_knight["name"]: red_knight["hp"],
-    }
+def battle(knights_config: dict) -> dict:
+    lancelot = preparation(knights_config)["Lancelot"]
+    mordred = preparation(knights_config)["Mordred"]
+    artur = preparation(knights_config)["Artur"]
+    red_knight = preparation(knights_config)["Red Knight"]
```
OleksandrPozden, line 97: No reason to store knights in Knights if this can be replaced with ```lancelot = Knight(some_params)

OleksandrPozden (changes requested):

Isles22:
Hi. Solution fixed

сб, 3 дек. 2022 г. в 14:26, Oleksandr Pozden ***@***.***>:

app/class_knight/knight.py
```diff
@@ -0,0 +1,28 @@
...
+    @staticmethod
+    def prepare(name: str, knights_config: dict) -> Knight:
+        knight_dict = knights_config[name]
+        current_knight = Knight(knight_dict["name"], knight_dict["power"],
+                                knight_dict["hp"], 0)
+        if len(knight_dict["armour"]) > 0:
+            for armour in knight_dict["armour"]:
+                current_knight.protection += armour["protection"]
+        current_knight.power += knight_dict["weapon"]["power"]
+        if knight_dict["potion"] is not None:
```
YuriiKindrat, line 22: u can write here just knight_dict["potion"] instead of "is not None"

app/class_knight/knight.py
```diff
@@ -0,0 +1,28 @@
...
+            current_knight.power += knight_dict["potion"]["effect"]["power"]
+            current_knight.hp += knight_dict["potion"]["effect"]["hp"]
+            if knight_dict["potion"]["effect"].get("protection") is not None:
```
YuriiKindrat, line 25: u can write here .get("protection", None)

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -86,129 +89,20 @@
...
+    red_knight = Knight.prepare("red_knight", knights_config)
+    lancelot.hp -= mordred.power - lancelot.protection
+    mordred.hp -= lancelot.power - mordred.protection
+    arthur.hp -= red_knight.power - arthur.protection
+    red_knight.hp -= arthur.power - red_knight.protection
+
+    return {lancelot.name: lancelot.hp if lancelot.hp > 0 else 0,
+            arthur.name: arthur.hp if arthur.hp > 0 else 0,
+            mordred.name: mordred.hp if mordred.hp > 0 else 0,
+            red_knight.name: red_knight.hp if red_knight.hp > 0 else 0}
```
Abnormaltype, line 105: you can write this logic in the class (if lancelot.hp > 0 else 0)

Abnormaltype (changes requested):

app/main.py
Isles22: thanks for the advice

app/class_knight/knight.py
```diff
@@ -0,0 +1,28 @@
+from __future__ import annotations
+
+
+class Knight:
+
+    def __init__(self, name: str, power: int, hp: int, protection: int) -> \
+            None:
+        self.name = name
```
OleksandrPozden, line 8: fix this
```suggestion
    def __init__(
             self, name: str, 
             power: int, hp: int, 
             protection: int
     ) -> None:
        self.name = name

```

OleksandrPozden (changes requested):

app/class_knight/knight.py
Isles22: yes, it's more readable

app/class_knight/knight.py
```diff
@@ -0,0 +1,31 @@
+from __future__ import annotations
+
+
+class Knight:
+
+    def __init__(
+            self, name: str,
+            power: int, hp: int,
+            protection: int
```
OleksandrPozden, line 9:
```suggestion
    def __init__(
            self, name: str,
            power: int,
            hp: int,
            protection: int

```

app/class_knight/knight.py
```diff
@@ -0,0 +1,31 @@
...
+        if len(knight_dict["armour"]) > 0:
+            for armour in knight_dict["armour"]:
+                current_knight.protection += armour["protection"]
+        current_knight.power += knight_dict["weapon"]["power"]
+        if knight_dict["potion"]:
+            current_knight.power += knight_dict["potion"]["effect"]["power"]
+            current_knight.hp += knight_dict["potion"]["effect"]["hp"]
+            if knight_dict["potion"]["effect"].get("protection", None):
+                current_knight.protection += \
+                    knight_dict["potion"]["effect"]["protection"]
```
OleksandrPozden, line 30:
```suggestion
        if knight_dict["potion"]:
            potion = knight_dict["potion"]["effect"]
            current_knight.power += potion["power"]
            current_knight.hp += potion["hp"]
            current_knight.protection += potion.get("protection",0)
```
can we have something like that?

OleksandrPozden (changes requested):

app/class_knight/knight.py
Isles22: I agree, it's much better

IvanRamyk (approved):
Well done!

PR #149, opened 2022-12-05
'Solution'

evgenijmartynuk07:


app/knights/apply_potion.py
```diff
@@ -0,0 +1,21 @@
...
+                if "power" in self.knight[hero]["potion"]["effect"]:
+                    self.knight[hero]["power"] += \
+                        self.knight[hero]["potion"]["effect"]["power"]
+
+                if "protection" in self.knight[hero]["potion"]["effect"]:
+                    self.knight[hero]["protection"] += \
+                        self.knight[hero]["potion"]["effect"]["protection"]
+
+                if "hp" in self.knight[hero]["potion"]["effect"]:
+                    self.knight[hero]["hp"] += \
```
OleksandrPozden, line 18: change this part. it look too cumbersome

OleksandrPozden (changes requested):

app/knights/apply_potion.py
```diff
@@ -0,0 +1,27 @@
...
+    def protection(self) -> dict:
+        if "protection" in self.knight["potion"]["effect"]:
+            self.knight["protection"] += \
+                self.knight["potion"]["effect"]["protection"]
+            return self.knight
+
+    def hp(self) -> dict:
+        if "hp" in self.knight["potion"]["effect"]:
+            self.knight["hp"] += self.knight["potion"]["effect"]["hp"]
+            return self.knight
```
Abnormaltype, line 27: you can write one method with loop for it

Abnormaltype (changes requested):
use the checklist before pushing your solution

app/knights/apply_potion.py
```diff
@@ -0,0 +1,21 @@
...
+        if "power" in self.knight["potion"]["effect"]:
+            self.knight["power"] += self.knight["potion"]["effect"]["power"]
+
+        if "protection" in self.knight["potion"]["effect"]:
+            self.knight["protection"] += \
+                self.knight["potion"]["effect"]["protection"]
+
+        if "hp" in self.knight["potion"]["effect"]:
+            self.knight["hp"] += self.knight["potion"]["effect"]["hp"]
+        return self.knight
```
Abnormaltype, line 21: you should create a loop and var stats = ("power", ...)

Abnormaltype (changes requested):

OleksandrPozden (changes requested):
I'm not too fond of it. Why do we have to hold self.knight as a dict in every single class?
Can't we have a different Knight class? And Potion and Armour etc. will receive a knight object as an argument in their methods

app/knights/knight.py
```diff
@@ -0,0 +1,30 @@
+class Knights:
```
IvanRamyk, line 1:
```suggestion
class Knight:

```

app/tournament/health.py
```diff
@@ -0,0 +1,14 @@
+from app.knights.knight import Knights
+
+
+class Health:
```
IvanRamyk, line 4: looks like you don't need a new class to perform such a check

app/tournament/round.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class Round:
+    def __init__(self, player1: Knights, player2: Knights) -> None:
+        self.player1 = player1
+        self.player2 = player2
+
+    def round(self) -> tuple:
+        self.player1.hp -= self.player2.power - self.player1.protection
+        self.player2.hp -= self.player1.power - self.player2.protection
```
IvanRamyk, line 11: i'd move health check to this method

IvanRamyk (changes requested):

app/tournament/round.py
```diff
@@ -0,0 +1,16 @@
...
+        if self.player1.hp < 0:
+            self.player1.hp = 0
+        if self.player2.hp < 0:
+            self.player2.hp = 0
+        return self.player1, self.player2
```
Serhii-Leonenko, line 16: You can use max(0, self.player.hp) to return zero if number is negative)

Serhii-Leonenko (approved):

PR #164, opened 2022-12-26
'Solution'

Evgen-Shepelenko:


app/main.py
```diff
@@ -86,129 +89,66 @@
...
+    )
+
+    red_knight = Knights(
+        name=knights_config["red_knight"]["name"],
+        power=knights_config["red_knight"]["power"],
+        hp=knights_config["red_knight"]["hp"],
+        armours=knights_config["red_knight"]["armour"],
+        weapon=knights_config["red_knight"]["weapon"],
+        potion=knights_config["red_knight"]["potion"]
+    )
```
IvanRamyk, line 128: all knights could be created in the loop

IvanRamyk (changes requested):

app/main.py
```diff
@@ -86,129 +89,42 @@
...
+            hp=knights_config[knight]["hp"],
+            armours=knights_config[knight]["armour"],
+            weapon=knights_config[knight]["weapon"],
+            potion=knights_config[knight]["potion"]
+        ))
 
-    # check if someone fell in battle
-    if lancelot["hp"] <= 0:
-        lancelot["hp"] = 0
+    list_knights[0].preparing_knight_battle()
```
IvanRamyk, line 105: why do not use a loop?

IvanRamyk (changes requested):

Abnormaltype (approved):
Well done!

PR #165, opened 2022-12-26
solution

the-andrew-007:


katevasilenko (changes requested):
I would prefer more clear variables with this `cfg`. All other is good

IvanRamyk (approved):

PR #168, opened 2022-12-28
Knights fighting

LavrykN:


app/potion.py
```diff
@@ -0,0 +1,13 @@
+from __future__ import annotations
+
+
+class Potion:
+    def __init__(self, name: str, effect: dict) -> None:
+        self.name = name
+        self.effect = effect
+
+    @staticmethod
```
olena-hul, line 9: The same comment about classmethods

app/weapon.py
```diff
@@ -0,0 +1,11 @@
+from __future__ import annotations
+
+
+class Weapon:
+    def __init__(self, name: str, power: int) -> None:
+        self.name = name
+        self.power = power
+
+    @staticmethod
```
olena-hul, line 9: And here

olena-hul (changes requested):

app/knight.py
```diff
@@ -48,12 +51,7 @@ def prepare_to_battle(self) -> Knight:
 
         if self.potion is not None:
             for name, score in self.potion.effect.items():
-                if name == "power":
-                    self.power += score
-                elif name == "hp":
-                    self.hp += score
-                elif name == "protection":
-                    self.protection += score
+                setattr(self, name, getattr(self, name) + score)
```
LavrykN, line 54: Not sure if is a good solution

IvanRamyk (approved):

PR #174, opened 2023-01-10
Solution

serg121192:


app/Ammunition/armours.py
```diff
@@ -0,0 +1,8 @@
+class Armour:
+
+    armours = {}
```
olena-hul, line 3: Why do you need to store armours in class attribute? Create Armour instances and add it to Knight instantly as it gets created

app/Heroes/knights.py
```diff
@@ -0,0 +1,46 @@
...
+            Knight.knights[self.name].armour.append(arm.part)
+            Knight.knights[self.name].protection += arm.protection
+
+    def potion_addition_stats(self, potion: Potion) -> None:
+        Knight.knights[self.name].power += Potion.potions[potion.name].power
+        Knight.knights[self.name].health += potion.health
+        Knight.knights[self.name].protection += potion.protection
+        Knight.knights[self.name].potion = potion.name
+
+    def geather_addition_stats(self, weapon: Weapon,
```
olena-hul, line 38: I didn't get the name of the variable

app/main.py
```diff
@@ -1,214 +1,86 @@
...
+from app.Heroes.knights import Knight
+from app.Ammunition.potions import Potion
+from app.Ammunition.weapons import Weapon
+
+
+def classes_creation(knights_config: dict) -> None:
+    for knight in knights_config:
+        Knight(knights_config[knight]["name"],
+               knights_config[knight]["power"],
+               knights_config[knight]["hp"])
```
olena-hul, line 11: Use knight directly, do not perform additional lookup in knights config

app/main.py
```diff
@@ -1,214 +1,86 @@
...
+            for eff in knights_config[knight]["potion"]["effect"]:
+                if eff == "power":
+                    _power = knights_config[knight][
+                        "potion"]["effect"]["power"]
+                elif eff == "hp":
+                    _health = knights_config[knight][
+                        "potion"]["effect"]["hp"]
+                elif eff == "protection":
+                    _protection = knights_config[knight][
+                        "potion"]["effect"]["protection"]
```
olena-hul, line 31: Do this in a loop using `setattr` and `getattr`

app/main.py
```diff
@@ -1,214 +1,86 @@
...
-
-    # apply weapon
-    mordred["power"] += mordred["weapon"]["power"]
+from app.Ammunition.armours import Armour
+from app.Heroes.knights import Knight
+from app.Ammunition.potions import Potion
+from app.Ammunition.weapons import Weapon
+
+
+def classes_creation(knights_config: dict) -> None:
```
olena-hul, line 7: Bad name

olena-hul:
The main big problem here is the attribute access. You are doing lots of extra operations
For example, instead of iterating through created objects, you still use input dict, get object by name from the dict
And when you need to change the object somehow, you are doing addditonal lookup by object's name however you already have the object
I didn't mention all such occurrences in comments, but you need to refactor all the code to make it simpler to understand and access to objects and their attributes should be direct and without additional lookup. It looks unreadable and adds extra complexity to the code

app/Ammunition/potions.py
```diff
@@ -0,0 +1,12 @@
+class Potion:
+
+    def __init__(self,
+                 name: str = None,
+                 power: int = 0,
+                 health: int = 0,
+                 protection: int = 0
+                 ) -> None:
```
olena-hul, line 8:
```suggestion
    def __init__(
        self,
        name: str = None,
        power: int = 0,
        health: int = 0,
        protection: int = 0
    ) -> None:

```

app/Heroes/knights.py
```diff
@@ -0,0 +1,35 @@
...
+                 potion: Potion | None,
+                 protection: int = 0
+                 ) -> None:
+        self.name = name
+        self.power = power + weapon.power + \
+            (potion.power if potion is not None else 0)
+        self.health = health + \
+            (potion.health if potion is not None else 0)
+        self.protection = protection + \
+            (potion.protection if potion is not None else 0)
```
olena-hul, line 25: This can be moved to a separate method. And you can iterate through ('power', 'health', 'protection') tuple and use `getattr` and `settattr` functions in order not to repeat the code

app/Heroes/knights.py
```diff
@@ -0,0 +1,35 @@
...
+        self.weapon = weapon.name
+        self.armour = []
+        if potion is None:
+            self.potion = None
```
olena-hul, line 29: This does not make sense

app/main.py
```diff
@@ -1,214 +1,52 @@
...
+    fighter_2.health -= fighter_1.power - fighter_2.protection
+    health_update(fighter_1)
+    health_update(fighter_2)
+
+
+def battle(knights_config: dict) -> dict:
+    lancelot = knights_creation_config(knights_config, "lancelot")
+    arthur = knights_creation_config(knights_config, "arthur")
+    mordred = knights_creation_config(knights_config, "mordred")
+    red_knight = knights_creation_config(knights_config, "red_knight")
```
olena-hul, line 42: You can iterate through all knights in dict and create Knight instances 
```suggestion
    knights = {}
    for knight in knights_config:
          knights[knight['name']] = create_knight(knight)
```

app/main.py
```diff
@@ -1,214 +1,52 @@
...
-
-    # apply weapon
-    mordred["power"] += mordred["weapon"]["power"]
+from app.Ammunition.armours import Armour
+from app.Heroes.knights import Knight
+from app.Ammunition.potions import Potion
+from app.Ammunition.weapons import Weapon
+
+
+def knights_creation_config(config: dict, knight: str) -> Knight:
```
olena-hul, line 7: Rename to `create_knight`

app/main.py
```diff
@@ -1,214 +1,52 @@
...
+    power = config[knight]["power"]
+    health = config[knight]["hp"]
+    weapon = Weapon(config[knight]["weapon"]["name"],
+                    config[knight]["weapon"]["power"])
+    armour = []
+    for _armour in config[knight]["armour"]:
+        armour.append(Armour(_armour["part"], _armour["protection"]))
+    potion = None
+    if config[knight]["potion"] is not None:
+        effect = config[knight]["potion"]["effect"]
```
olena-hul, line 18: store potion in variable and reuse it

app/Heroes/knights.py
```diff
@@ -0,0 +1,35 @@
...
+        else:
+            self.potion = potion.name
+        if len(armour) != 0:
```
olena-hul, line 32: This check is not needed

olena-hul:
Good job!

olena-hul (approved):

PR #190, opened 2023-02-02
Solution

Roman28101:


app/fight.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class Arena:
+
+    @staticmethod
+    def battle(knights: list[Knight]) -> None:
+        knight1 = Knight
+        knight2 = Knight
+        knight3 = Knight
+        knight4 = Knight
```
katevasilenko, line 11: This part kind of mess me. Don’t create objects of class without parameters. Also, it would be better to name variables by knight’s name.

app/main.py
```diff
@@ -1,3 +1,27 @@
+from app.knights import Knight
+from app.fight import Arena
+
+
+def battle(knights: dict) -> dict:
```
katevasilenko, line 5: It is not a good idea to name two functions of method in the same name in such small project. I would prefer to rename one of those

katevasilenko: You can create objects of knight here directly from the dict and put them as arguments to any functions and methods you need

app/main.py
```diff
@@ -1,3 +1,27 @@
...
+    Arena.battle(list_knights)
+    knights_dict = {}
+    for knight in list_knights:
+        if knight.hp <= 0:
+            knight.hp = 0
+        knights_dict[knight.name] = knight.hp
+    return knights_dict
+
+
 KNIGHTS = {
```
katevasilenko, line 25: Please put this dict to separate module

app/main.py
```diff
@@ -1,3 +1,27 @@
...
+def battle(knights: dict) -> dict:
+    list_knights = []
+    for key, value in knights.items():
+        knight = Knight(value["name"], value["power"], value["hp"])
+        knight.use_weapon(value["weapon"]["power"])
+        if value["armour"] is not None:
+            knight.use_armour(value["armour"])
+        if value["potion"] is not None:
+            knight.use_potion(value["potion"]["effect"])
+        list_knights.append(knight)
```
katevasilenko, line 14: Please remove this logic to separate function

app/fight.py
```diff
@@ -0,0 +1,24 @@
...
+            if knight.name == "Artur":
+                knight2 = knight
+            if knight.name == "Mordred":
+                knight3 = knight
+            if knight.name == "Red Knight":
+                knight4 = knight
+        knight1.hp -= knight3.power
+        knight3.hp -= knight1.power
+        knight2.hp -= knight4.power
+        knight4.hp -= knight2.power
```
katevasilenko, line 24: Please name variables after knight names. This are bad names

katevasilenko (changes requested):

Vasyl-Smutok (changes requested):
@Roman28101  please do as @katevasilenko  said

app/knights.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def use_armour(self, armour: list) -> None:
+        for item in armour:
+            self.hp += item["protection"]
+
+    def use_potion(self, effects: dict) -> None:
+        for key, value in effects.items():
+            if key == "power":
+                self.power += value
+            if key == "protection":
```
Grayder0152, line 19: You can use `or` to avoid next `if`

app/fight.py
```diff
@@ -0,0 +1,22 @@
...
+            knight = Knight(value["name"], value["power"],
+                            value["hp"])
+            knight.use_weapon(value["weapon"]["power"])
+            if value["armour"] is not None:
+                knight.use_armour(value["armour"])
+            if value["potion"] is not None:
+                knight.use_potion(value["potion"]["effect"])
+            dict_knights[key] = knight
+
+        dict_knights["lancelot"].hp -= dict_knights["mordred"].power
```
Grayder0152, line 18: You can create method in Arena witch get names of opponents and do 18 and 19 line (to aboid dublication)

app/fight.py
```diff
@@ -0,0 +1,22 @@
+from app.knights import Knight
+
+
+class Arena:
+    @staticmethod
+    def opposition(knights: dict) -> dict:
+        dict_knights = {}
```
Grayder0152, line 7: Create this variable as class or instance variable

app/knights.py
```diff
@@ -0,0 +1,22 @@
+class Knight:
+
+    def __init__(self, name: str, power: int, hp: int) -> None:
+        self.name = name
+        self.power = power
+        self.hp = hp
```
Grayder0152, line 6: Create `hp` and `power` as protected attributes, add for them property and set limit for `hp` (if hp < 0 ... to avoid it in `battle` function)

app/main.py
```diff
@@ -1,3 +1,15 @@
+from app.fight import Arena
+
+
+def battle(knighs: dict) -> dict:
+    result_dict = {}
+    for knight in Arena.opposition(knighs).values():
```
Grayder0152, line 6: it is not necessary but you can use dict comprehension

app/fight.py
```diff
@@ -0,0 +1,22 @@
...
+        for key, value in knights.items():
+            knight = Knight(value["name"], value["power"],
+                            value["hp"])
+            knight.use_weapon(value["weapon"]["power"])
+            if value["armour"] is not None:
```
Grayder0152, line 12: I recommend to read about `:=` in Python, cool operator, but it is not necessary

Grayder0152 (changes requested):

PR #196, opened 2023-02-06
Solution

manjustice:


app/main.py
```diff
@@ -1,214 +1,40 @@
...
-    }
-}
+from app.items import Weapon, Armour, Potion
+from app.warriors import Knight
+from app.data import KNIGHTS
 
 
-def battle(knightsConfig):
-    # BATTLE PREPARATIONS:
+def battle(knights_config: dict) -> dict:
```
bihunva, line 6: you can use Dict from typing module to specify what type of keys and values will be returned in the dictionary

app/main.py
```diff
@@ -1,214 +1,40 @@
...
-    lancelot["power"] += lancelot["weapon"]["power"]
+        for armour in config["armour"]:
+            knight.armour_up(Armour(part=armour["part"],
+                                    protection=armour["protection"]))
 
-    # apply potion if exist
-    if lancelot["potion"] is not None:
-        if "power" in lancelot["potion"]["effect"]:
-            lancelot["power"] += lancelot["potion"]["effect"]["power"]
+        if config["potion"]:
```
bihunva, line 21: It seems better to transfer all the logic of working with the "potion" to the "use_potion" method of the "Knight" class. And in the main file just call them.

bihunva (changes requested):

AndereLion (approved):

app/main.py
```diff
@@ -1,214 +1,37 @@
...
-            lancelot["power"] += lancelot["potion"]["effect"]["power"]
+        for armour in config["armour"]:
+            knight.armour_up(Armour(part=armour["part"],
+                                    protection=armour["protection"]))
 
-        if "protection" in lancelot["potion"]["effect"]:
-            lancelot["protection"] += lancelot["potion"]["effect"]["protection"]
+        if config["potion"]:
+            knight.use_potion(Potion(name=config["potion"]["name"],
+                                     effects=config["potion"]["effect"]))
```
YuriiKindrat, line 25: u can call this methods inside init so all config will be added authomaticaly

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,214 +1,27 @@
...
+                        armour=config["armour"],
+                        potion=config["potion"])
+        knights[warrior] = knight
 
-    # apply weapon
-    lancelot["power"] += lancelot["weapon"]["power"]
+    knights["lancelot"].strike_enemy(knights["mordred"])
+    knights["mordred"].strike_enemy(knights["lancelot"])
+    knights["arthur"].strike_enemy(knights["red_knight"])
+    knights["red_knight"].strike_enemy(knights["arthur"])
```
YuriiKindrat, line 22: try to rewrite this method, so there won't need to call it for every knight
U can make this method work for pair

YuriiKindrat (changes requested):

Nattalli (approved):
Good job!

PR #197, opened 2023-02-07
Solution

MaybeTI:


app/Knights/Preparation.py
```diff
@@ -0,0 +1,29 @@
+class PrepareKnight:
```
manjustice, line 1: Rename this class to Knight because it's not proper way to name class using verb. And also rename package Knights and modules BATTLE, KNIGHTS, Preparation to lowercase

manjustice (changes requested):

SafonovVladimir:
Привіт. 
Я думаю, що class - це якась сутність (наприклад knight, armour, weapon або potion).
А підготовка до бою та сам бій - це вже дія, або функція.
Я би зробив взагалі окремо класи під кожну з цих сутністей.
Як вважаєте?
PS. В принципі бій також можна розглядати як клас, но це вже хто як "бачить".

manjustice (approved):
GJ!

app/main.py
```diff
@@ -1,214 +1,22 @@
...
+def battle(knights_config: dict) -> dict:
+    for knight, info in knights_config.items():
+        knight = Knight(name=info["name"],
+                        power=info["power"],
+                        hp=info["hp"])
+        Knight.knights_dict[info["name"]] = knight
+        knight_update = Knight.knights_dict[info["name"]]
+        knight_update.update(armour=info["armour"],
+                             weapon=info["weapon"],
+                             potion=info["potion"])
```
YuriiKindrat, line 14: u can make this in init method

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -3,15 +3,7 @@
...
-    for knight, info in knights_config.items():
-        knight = Knight(name=info["name"],
-                        power=info["power"],
-                        hp=info["hp"])
-        Knight.knights_dict[info["name"]] = knight
-        knight_update = Knight.knights_dict[info["name"]]
-        knight_update.update(armour=info["armour"],
-                             weapon=info["weapon"],
-                             potion=info["potion"])
+    Knight.upgrade(knights_config=knights_config)
```
YuriiKindrat, line 6: no need to kall this method here, u can just call self.update in init method

YuriiKindrat (changes requested):

app/main.py
MaybeTI: > no need to kall this method here, u can just call self.update in init method
@YuriiKindrat 
I don't understand what needs to be done, I was at the Q&A and the mentor told me that he also doesn't understand what needs to be done, so can you explain

YuriiKindrat: @MaybeTI u have 2 similar names for methods, so i was confused

app/Knights/preparation.py
```diff
@@ -0,0 +1,41 @@
...
+        self.apply_weapon(weapon)
+        self.apply_potion(potion)
+
+    @staticmethod
+    def upgrade(knights_config: dict) -> None:
+        for knight, info in knights_config.items():
+            knight = Knight(name=info["name"],
+                            power=info["power"],
+                            hp=info["hp"])
+            Knight.knights_dict[info["name"]] = knight
```
YuriiKindrat, line 37: u can make this inside ur init method, like Knight.knights_dict[self.name] = self

app/Knights/preparation.py
```diff
@@ -0,0 +1,41 @@
...
+            knight_update = Knight.knights_dict[info["name"]]
```
YuriiKindrat, line 38: also, i don't understand why should u create a new variable for the same knight, u already have variable knight with the same obj

app/Knights/preparation.py
```diff
@@ -0,0 +1,41 @@
...
+            knight_update.update(armour=info["armour"],
+                                 weapon=info["weapon"],
+                                 potion=info["potion"])
```
YuriiKindrat, line 41: u can add this parameters to an init arguments and call method inside ur method init, i've talked about this method

app/Knights/preparation.py
```diff
@@ -0,0 +1,41 @@
...
+            if "hp" in potion["effect"]:
+                self.hp += potion["effect"]["hp"]
+
+    def update(self, armour: list, weapon: dict, potion: dict) -> None:
+        self.protection_update(armour)
+        self.apply_weapon(weapon)
+        self.apply_potion(potion)
+
+    @staticmethod
+    def upgrade(knights_config: dict) -> None:
```
YuriiKindrat, line 32: also make a more descriptive name for this method, it has the similar name as method above)

YuriiKindrat (changes requested):

PR #203, opened 2023-02-08
Solution

WDemidenko:


app/main.py
```diff
@@ -1,214 +1,20 @@
...
+from Knights_data.Knights import Knight
+from Knights_data.unformated import KNIGHTS
+
+
+def battle(knightsconfig: dict) -> dict:
+    knights = {}
+    for stats in knightsconfig.values():
+        knight = KnightStats(stats["name"], stats["power"], stats["hp"],
+                             stats["armour"], stats["weapon"], stats["potion"])
+        KnightStats.potion(KnightStats.knights[stats["name"]])
```
kuznetsb, line 11: Maybe it' possible to call this method potion inside KnightStats class while creating object of it.

app/main.py
```diff
@@ -1,214 +1,20 @@
...
+        knights[stats["name"]] = Knight(knight)
+    knightsconfig = knights
```
kuznetsb, line 13: I think you don't need to create new variable here and can use just knights

kuznetsb (approved):
Cool

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #206, opened 2023-02-09
Solution

usrofgh:


app/knight.py
```diff
@@ -0,0 +1,88 @@
...
+            pass
+
+    def hit(self, other: Knight) -> None:
+        if self.__power >= other.__protection + other.__hp:
+            other.__protection = 0
+            other.__hp = 0
+
+        elif self.__power >= other.__protection:
+            other.__hp -= self.__power - other.__protection
+            other.__protection = 0
```
amber-marichi, line 84: I really like your classes, so why dont you use all those getters here in this method? :)

app/main.py
```diff
@@ -1,214 +1,51 @@
...
-    arthur["power"] += arthur["weapon"]["power"]
+    lancelot.hit(mordred)
+    mordred.hit(lancelot)
 
-    # apply potion if exist
-    if arthur["potion"] is not None:
-        if "power" in arthur["potion"]["effect"]:
-            arthur["power"] += arthur["potion"]["effect"]["power"]
+    arthur.hit(red_knight)
+    red_knight.hit(arthur)
```
amber-marichi, line 45: not sure about all this being so hard-coded here. what about if we wanted to see eg Red Knight vs Mordred? maybe this can be revised a bit

amber-marichi (changes requested):
very nice actually. few improvements maybe

app/knight.py
usrofgh: i'm not sure that we need to use getters out of a class, cause we have access in a class to variables directly. I've not readed about it yet, so if you have some information/article which refutes my approach of code. I'll read it with great plesaure

app/main.py
usrofgh: done

app/knight.py
amber-marichi: in fact i just wanted to point out that it seemed strange that after you've implemented in Knight methods like "get_armour", "get_hp", "get_power" etc they havent been used here. I mean it is ok to address attributes on self but it would be cool to get attribute of other via those methods. Anyway that was more like suggestion so dont mind if you dont feel like it.

app/main.py
```diff
@@ -30,19 +30,20 @@ def create_knights(knights: dict) -> {Knight}:
...
-
-    lancelot.hit(mordred)
-    mordred.hit(lancelot)
-
-    arthur.hit(red_knight)
-    red_knight.hit(arthur)
+    participants = {
+        knights["Lancelot"]: knights["Mordred"],
+        knights["Artur"]: knights["Red Knight"]
+    }
```
amber-marichi, line 44: these methods look great, but the names of participants are still hard-coded inside battle method. How about making it receive these names say in form of list of tuples(or dicts or else) so in case when you need to change dueling pairs you just pass different tuple and dont have to rewrite anything in battle method?

amber-marichi (changes requested):
it is better but i can still see potential upgrade in battle

app/knight.py
usrofgh: It seems to me it isn't a bad idea leave getters there. Now they are unnecessary. But suppose that in a real project someone want in the future get name/power etc of a knight. Without getters he'll need to realize them. such methods like get_name/get_hp/get_power will use often I think. It's like an example with battle method. You've wrote about potential necessity to change opponents.  According to task description we don't need to change them. But in future can, here the same

Also noticed that IDLE doesn't swear that I don't use them

app/main.py
usrofgh: I get a dict from ```create_knights``` method in next structure: ```{name_of_knight: knight_object}```
after I get objects by name in ```participants```
as I understand do I need to change ```participants``` variable which need to have just 2 tuple of names(str)?

amber-marichi: i would suggest just to add another parameter to battle() func after knights_config, something that you could iterate through like "(knight_one, knight_two)" so that another dueling pair would be taken just by passing these variables ( " knights[knight_one]: knights[knight_two] ") and so you wouldnt need to have any name like "Lancelot" and other inside battle() at all

app/knight.py
amber-marichi: oh no, i havent asked you to remove getters :) i like them too , i just suggested that you could have used them to get the necessary info from the inside of instance passed into hit() method. For example:    if self.__power >= other.get_power() + other.get_hp()  -- like that.

usrofgh: I just don't understand why do we need to use  ```self``` like ```self.__power``` and ```other``` like ```other.get_power()```
We use getters outside cause it isn't recommended touch protected/private attributes/methods directly
But in this example ```self``` and ```other``` inside of their class and python allows us to use them directly
So why do we need  touch the attribute of ```self``` directly, and an attribute of ```other``` through getter? Maybe do you have any example where it's shown, for familiarization and analyzing

amber-marichi: im sorry i dont have any example. I get your point so i wont insist on changing this part here. I guess its just something i would do but that doesnt mean it is the only correct way.

amber-marichi (approved):

app/main.py
```diff
@@ -1,214 +1,52 @@
...
-
-    # Return battle results:
-    return {
-        lancelot["name"]: lancelot["hp"],
-        arthur["name"]: arthur["hp"],
-        mordred["name"]: mordred["hp"],
-        red_knight["name"]: red_knight["hp"],
+def battle(knights_config: dict) -> {}:
+    knights = create_knights(knights_config)
+    [knight.prepare_for_battle() for knight in knights.values()]
```
YuriiKindrat, line 40: u can use this method inside init method

app/main.py
```diff
@@ -1,214 +1,52 @@
...
+        name = params["name"]
+        power = params["power"]
+        hp = params["hp"]
 
-    # apply armour
-    mordred["protection"] = 0
-    for a in mordred["armour"]:
-        mordred["protection"] += a["protection"]
+        armour = [Armour(arm["part"], arm["protection"])
+                  for arm in params["armour"]]
```
YuriiKindrat, line 18: arm is a bad name for variable

YuriiKindrat (changes requested):

app/knight.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def _apply_potion(self) -> None:
+        effects = self._potion.effect
+        # print(effects)
+        if "power" in effects:
+            self._power += effects["power"]
+        if "hp" in effects:
+            self._hp += effects["hp"]
+        if "protection" in effects:
+            self._protection += effects["protection"]
```
Serhii-Leonenko, line 72: Read the checklist:

image


Hint: use `setattr` for this purpose

Serhii-Leonenko (changes requested):

app/main.py
```diff
@@ -1,214 +1,51 @@
...
+            potion = Potion(params["potion"]["name"],
+                            params["potion"]["effect"])
+        except TypeError:
+            potion = None
 
-    # apply potion if exist
-    if mordred["potion"] is not None:
-        if "power" in mordred["potion"]["effect"]:
-            mordred["power"] += mordred["potion"]["effect"]["power"]
+        knights_dict[name] = Knight(name, power, hp, armour, weapon, potion)
```
YuriiKindrat, line 28: u can use class attribute here, with all knights, or just delete it from ur class, cause u don't use it anywhere

YuriiKindrat (changes requested):

app/main.py
usrofgh: As I understand you talk about ```_all_knights``` class attribute, I use it in class method ```get_hp_all_knights```
If I delete ```knights_dict[name]``` here and will just return class attribute, then in ```battle``` function i'll need to create loop with condition, cause i must specify knights each other by their name
Do I need to do so or somewhat differently ?

YuriiKindrat: u didn't fix that

YuriiKindrat (changes requested):

PR #209, opened 2023-02-09
Solution

ZhAlexR:


app/main.py
```diff
@@ -1,214 +1,21 @@
...
+from app.data import KNIGHTS
+from app.battle_implement.implementation import create_knight
+from app.battle_implement.battle_process import battle_process, battles_result
 
 
-def battle(knightsConfig):
-    # BATTLE PREPARATIONS:
+def battle(knights_dict: dict) -> dict:
+    knights = [create_knight(knights_dict.get(knight))
+               for knight in knights_dict]
```
Serhii-Leonenko, line 8: you iterate through knight_dict, for what you need `knights_dict.get(knight)`, use dict.items() or .values()

app/main.py
```diff
@@ -1,214 +1,21 @@
...
 
-    # lancelot
-    lancelot = knightsConfig["lancelot"]
+    lancelot = knights[0]
+    arthur = knights[1]
+    mordred = knights[2]
+    red_knight = knights[3]
+    battle_process(lancelot, mordred)
+    battle_process(arthur, red_knight)
```
Serhii-Leonenko, line 15: I think it's better to store created knights in the dict where keys will be names and values will be objects

app/battle_implement/implementation.py
```diff
@@ -0,0 +1,42 @@
...
+
+from app.entities.knight import Knight
+from app.entities.stuff import Weapon, ArmourPart, SetArmourScope, Potion
+
+
+def crete_armour(armours: list) -> SetArmourScope | list:
+    if armours:
+        armour_parts = [ArmourPart(armour.get("part"),
+                                   armour.get("protection"))
+                        for armour in armours]
```
Serhii-Leonenko, line 11: fix indentation here, it has bad readability. Use something like this

image

app/entities/stuff.py
```diff
@@ -0,0 +1,46 @@
...
+class SetArmourScope:
+
+    def __init__(self, armour_parts: list[ArmourPart]) -> None:
+        self.armour_parts = armour_parts
+        self.protection_level = 0
+
+    def calc_protection_level(self) -> int:
+        if self.armour_parts:
+            return sum([armour.protection_level
+                        for armour in self.armour_parts])
```
Serhii-Leonenko, line 22: Use sum without `[]`

app/battle_implement/implementation.py
```diff
@@ -0,0 +1,42 @@
...
+    power = knight_info.get("power")
+    hp = knight_info.get("hp")
+    armour_scope = crete_armour(knight_info.get("armour"))
+    weapon = create_weapon(knight_info.get("weapon"))
+    potion = create_potion(knight_info.get("potion"))
+    knight = Knight(name=name,
+                    power=power, hp=hp,
+                    armour=armour_scope,
+                    weapon=weapon,
+                    potion=potion)
```
Serhii-Leonenko, line 41: Fix indentaition here

Serhii-Leonenko (changes requested):

app/battle_implement/implementation.py
```diff
@@ -0,0 +1,48 @@
...
+    if armours:
+        armour_parts = [
+            ArmourPart(
+                armour.get("part"), armour.get("protection")
+            )
+            for armour in armours
+        ]
+        armour_scope = SetArmourScope(armour_parts)
+        return armour_scope
+    return SetArmourScope([])
```
Serhii-Leonenko, line 17: Using a default value for the armour_parts argument in the init of a class is better than putting an empty list like this` SetArmourScope([])`. Also, I think it's better to change the name of the class, maybe just ArmourScoupe without Set

Serhii-Leonenko (changes requested):

app/main.py
ZhAlexR: Thanks, it looks really beater!

app/battle_implement/implementation.py
ZhAlexR: Thanks, it looks really beater!

app/battle_implement/battle_process.py
```diff
@@ -0,0 +1,15 @@
+from __future__ import annotations
+
+from app.entities.knight import Knight
+
+
+def battle_process(first_knight: Knight, second_knight: Knight) -> None:
+    first_knight.battle_preparation()
+    second_knight.battle_preparation()
```
YuriiKindrat, line 8: u call this methods inside ur init method

app/main.py
```diff
@@ -1,214 +1,25 @@
...
-        mordred["hp"] = 0
-
-    # 2 Arthur vs Red Knight:
-    arthur["hp"] -= red_knight["power"] - arthur["protection"]
-    red_knight["hp"] -= arthur["power"] - red_knight["protection"]
+def battle(knights_dict: dict) -> dict:
+    knights = {
+        knight.get("name"): create_knight(knight)
+        for knight
+        in knights_dict.values()
```
YuriiKindrat, line 10: u can use items() method here, cause name is saved as a key of knights_dict

YuriiKindrat (changes requested):

app/main.py
ZhAlexR: @YuriiKindrat  Ok, but we have different names of Arthur as a key it is "arthur" as a value inside dict it is "Artur" without "h", that why i decided to use only dict.values()

YuriiKindrat (approved):

PR #217, opened 2023-02-13
'Solution'

dominyevhen:


app/actions/battle.py
```diff
@@ -0,0 +1,15 @@
+def action(rivals: dict, first: str, second: str) -> dict:
+    # calculation of the knight's hp
+    hp_first = rivals[second]["power"] - rivals[first]["protection"]
+    rivals[first]["hp"] -= hp_first
+    hp_second = rivals[first]["power"] - rivals[second]["protection"]
+    rivals[second]["hp"] -= hp_second
```
Bohdan-Salamakha, line 6: 1. use get method instead of parenthesis everywhere to avoid KeyError
rivals.get(second, {}).get("hp")
2.  create separated variables for rivals.get(first) and rivals.get(second) to not dublicate code

app/actions/battle.py
```diff
@@ -0,0 +1,15 @@
+def action(rivals: dict, first: str, second: str) -> dict:
+    # calculation of the knight's hp
+    hp_first = rivals[second]["power"] - rivals[first]["protection"]
+    rivals[first]["hp"] -= hp_first
+    hp_second = rivals[first]["power"] - rivals[second]["protection"]
+    rivals[second]["hp"] -= hp_second
+
+    # check if someone fell in battle
+    if rivals[first]["hp"] <= 0:
```
Bohdan-Salamakha, line 9: just < 0
u don't need to set this value to zero if it's zero already

app/actions/battle.py
```diff
@@ -0,0 +1,15 @@
...
+    hp_first = rivals[second]["power"] - rivals[first]["protection"]
+    rivals[first]["hp"] -= hp_first
+    hp_second = rivals[first]["power"] - rivals[second]["protection"]
+    rivals[second]["hp"] -= hp_second
+
+    # check if someone fell in battle
+    if rivals[first]["hp"] <= 0:
+        rivals[first]["hp"] = 0
+
+    if rivals[second]["hp"] <= 0:
```
Bohdan-Salamakha, line 12: just < 0
u don't need to set this value to zero if it's zero already

app/actions/preparations.py
```diff
@@ -0,0 +1,22 @@
+def battle_preparations(knights_config: dict) -> dict:
+    new_knights = knights_config
```
Bohdan-Salamakha, line 2: use copy mehod/func to copy not just a link but and data for mutable data types

app/knights_list.py
```diff
@@ -0,0 +1,98 @@
...
+            "effect": {
+                "hp": +10,
+                "power": +5,
+            }
+        }
+    }
+}
+# creation of knights
+for key, value in knights.items():
+    Knights(key, value)
```
Bohdan-Salamakha, line 98: what is it?
create a function for this logic and it's super necessary to call it here use startpoint if __name__ == "__main__":...

Bohdan-Salamakha (changes requested):
v1

ArsenPidhoretskyi (changes requested):
Create class Knight and use it instead of dict

PR #220, opened 2023-02-19
Solut

neagarr:


neagarr:
@Abnormaltype

app/main.py
```diff
@@ -1,214 +1,36 @@
...
-    lancelot = knightsConfig["lancelot"]
+    for key, value in knights.items():
+        locals()[key] = Fighters(
+            value["name"],
+            value["power"],
+            value["hp"],
+            value["armour"],
+            value["weapon"],
+            value["potion"]
+        )
```
ArsenPidhoretskyi, line 18: NO! It is too bad to do like this
Create another way

app/main.py
```diff
@@ -1,214 +1,36 @@
...
-            lancelot["protection"] += lancelot["potion"]["effect"]["protection"]
+        for fighter_2 in fighters_list:
+            if fighter_1.name == "Lancelot" and fighter_2.name == "Mordred":
+                Battle.fight(fighter_1, fighter_2)
+            elif fighter_1.name == "Artur" and fighter_2.name == "Red Knight":
+                Battle.fight(fighter_1, fighter_2)
 
-        if "hp" in lancelot["potion"]["effect"]:
-            lancelot["hp"] += lancelot["potion"]["effect"]["hp"]
+        result[fighter_1.name] = fighter_1.hp
```
ArsenPidhoretskyi, line 34: No need cycle write inline this

ArsenPidhoretskyi: We already know about who will fight with whom

app/arena/preparation.py
```diff
@@ -0,0 +1,24 @@
...
+    def potion_prepare(fighter: Fighters) -> None:
+        if fighter.potion is not None:
+            if "power" in fighter.potion["effect"]:
+                fighter.power += fighter.potion["effect"]["power"]
+
+            if "protection" in fighter.potion["effect"]:
+                fighter.protection += fighter.potion["effect"]["protection"]
+
+            if "hp" in fighter.potion["effect"]:
+                fighter.hp += fighter.potion["effect"]["hp"]
```
ArsenPidhoretskyi, line 24: Use cycle here with getattr and setattr

ArsenPidhoretskyi (changes requested):

neagarr (commented):
In the README.md the name is "Arthur" and in the test_main.py it is "Artur". This may cause some difficulty in getting the job done correctly.

app/arena/preparation.py
```diff
@@ -0,0 +1,25 @@
+from app.hall.fighters import Fighters
+
+
+class Prepare:
+    @staticmethod
+    def fighter_preparation(fighter: Fighters) -> None:
+        Prepare.armour_prepare(fighter)
+        Prepare.weapon_prepare(fighter)
+        Prepare.potion_prepare(fighter)
```
Grayder0152, line 9:
```suggestion
    def fighter_preparation(self, fighter: Fighters) -> None:
        self.armour_prepare(fighter)
        self.weapon_prepare(fighter)
        self.potion_prepare(fighter)
```

app/main.py
```diff
@@ -1,214 +1,21 @@
...
-            "effect": {
-                "hp": +10,
-                "power": +5,
-            }
-        }
-    }
-}
+from app.hall.fighters import Fighters
+from app.arena.preparation import Prepare
+from app.arena.fight import Battle
```
Grayder0152, line 3: Pls, sort imports

app/main.py
```diff
@@ -1,214 +1,21 @@
...
-def battle(knightsConfig):
-    # BATTLE PREPARATIONS:
+def battle(knights: dict) -> dict:
+    fighters_list = []
+    battle_result = {}
 
-    # lancelot
-    lancelot = knightsConfig["lancelot"]
+    for key, value in knights.items():
+        globals()[key] = Fighters(value)
```
Grayder0152, line 11: Why do you use `globals`?

app/hall/fighters.py
```diff
@@ -0,0 +1,8 @@
+# from __future__ import annotations
+
+
+class Fighters(object):
+    def __init__(self, fighter: dict) -> None:
+        for key in fighter:
+            setattr(self, key, fighter[key])
```
Grayder0152, line 7: is it not clear logic, pls add arguments

Grayder0152 (changes requested):

app/main.py
neagarr: I used "globals()[key]" or "locals()[key]" because when using just "key" in a loop, one instance of the class was created and it was just overwritten with each iteration for some reason. Global() or local() allowed to create a new instance of the class at each iteration. But now it is ok

app/hall/fighters.py
neagarr: Pls, give me a hint, because in the last commit I had this problem required several times as many lines of code

ArsenPidhoretskyi: You need to set only specific key:
hp, protection, power

app/main.py
```diff
@@ -1,214 +1,21 @@
...
 
-def battle(knightsConfig):
-    # BATTLE PREPARATIONS:
+def battle(knights: dict) -> dict:
+    fighters_list = []
+    battle_result = {}
 
-    # lancelot
-    lancelot = knightsConfig["lancelot"]
+    for key, value in knights.items():
```
ArsenPidhoretskyi, line 9: key, value is not clear variable names

app/main.py
```diff
@@ -1,214 +1,21 @@
...
-    for a in lancelot["armour"]:
-        lancelot["protection"] += a["protection"]
+        key = Prepare(value)
+        key.fighter_preparation()
+        fighters_list.append(key)
 
-    # apply weapon
-    lancelot["power"] += lancelot["weapon"]["power"]
+    Battle.fight(fighters_list[0], fighters_list[2])
+    Battle.fight(fighters_list[1], fighters_list[3])
```
ArsenPidhoretskyi, line 16: Not so clear
Use dict instead of list

app/main.py
```diff
@@ -1,214 +1,21 @@
...
 
-    # apply potion if exist
-    if lancelot["potion"] is not None:
-        if "power" in lancelot["potion"]["effect"]:
-            lancelot["power"] += lancelot["potion"]["effect"]["power"]
+    for i in range(len(fighters_list)):
```
ArsenPidhoretskyi, line 18: why you need an index you can iterate through values

ArsenPidhoretskyi (changes requested):

app/main.py
neagarr: Yes, thanks

app/main.py
neagarr: 1. Hp, protection and power are not enough, because later in the code I will need more instance attributes such as protection, potion, name..
2. Key, value is not clear variable names. But why I should use clear variable names?

app/main.py
neagarr: don't understand how can I do it better with dict... no ideas... maybe hint?))

app/hall/fighters.py
ArsenPidhoretskyi: @neagarr not fixed
add clear logic to init

app/main.py
ArsenPidhoretskyi: key will be name of Knight and value obj of Knight
Using indexes for fight is worst solution

ArsenPidhoretskyi (changes requested):

PR #231, opened 2023-03-02
Develop

VladyslavKaminskyi:


app/items.py
```diff
@@ -0,0 +1,25 @@
...
+        self.protection_effect = None
+        self.power_effect = None
+        self.hp_effect = None
+        for key, value in potion_conf["effect"].items():
+            if key == "hp":
+                self.hp_effect = value
+            if key == "power":
+                self.power_effect = value
+            if key == "protection":
+                self.protection_effect = value
```
ArsenPidhoretskyi, line 19: Use loop instead of duplicate code
you can use getattr, setattr
Also key, value so bad names

app/knight.py
```diff
@@ -0,0 +1,73 @@
...
+            print(f"{self.name} drink {self.potion.name} and get:")
+            if self.potion.hp_effect is not None:
+                self.hp += self.potion.hp_effect
+                print(f"{self.potion.hp_effect} HP")
+            if self.potion.power_effect is not None:
+                self.power += self.potion.power_effect
+                print(f"{self.potion.power_effect} power")
+            if self.potion.protection_effect is not None:
+                self.protection += self.potion.protection_effect
+                print(f"{self.potion.protection_effect} protection.")
```
ArsenPidhoretskyi, line 39: Again code duplicate

ArsenPidhoretskyi (changes requested):

PR #236, opened 2023-03-14
Solution

AlexandrZhezheria:


app/main.py
```diff
@@ -1,214 +1,36 @@
...
 
     # 2 Arthur vs Red Knight:
-    arthur["hp"] -= red_knight["power"] - arthur["protection"]
-    red_knight["hp"] -= arthur["power"] - red_knight["protection"]
+    members["arthur"]["hp"] -= (
+        members["red_knight"]["power"] - members["arthur"]["protection"]
+    )
+    members["red_knight"]["hp"] -= (
+        members["arthur"]["power"] - members["red_knight"]["protection"]
+    )
```
ArsenPidhoretskyi, line 22: You can move this code to function which accepts two knight

ArsenPidhoretskyi (changes requested):

app/main.py
AlexandrZhezheria: Вы хотите чтобы я написал функцию  и вынес туда этот код, которая принимает рандомно 2 рыцаря перед боем? Но у нас же в условии есть какие рыцари кто и с кем будут биться конкретно?

katevasilenko: For sure, but you have some code repeating right now. And imagine you add 10 new knights, your code will grow to much. Separate function helps you avoid this possible issue

katevasilenko (changes requested):

app/main.py
```diff
@@ -1,213 +1,25 @@
...
-    # Return battle results:
     return {
-        lancelot["name"]: lancelot["hp"],
-        arthur["name"]: arthur["hp"],
-        mordred["name"]: mordred["hp"],
-        red_knight["name"]: red_knight["hp"],
+        "Lancelot": fight_1["Lancelot"],
+        "Artur": fight_2["Artur"],
+        "Mordred": fight_1["Mordred"],
+        "Red Knight": fight_2["Red Knight"],
```
ArsenPidhoretskyi, line 22: You can simplify this)
```suggestion
        **fight_1,
        **fight_2
```

app/main.py
```diff
@@ -1,213 +1,25 @@
...
+        knight = Knight(value["name"], value["power"], value["hp"])
+        knight.use_weapon(value["weapon"]["power"])
+        knight.use_armour(value["armour"])
+        knight.use_potion(value["potion"])
+        list_knights.append(knight)
 
-    if red_knight["hp"] <= 0:
-        red_knight["hp"] = 0
+    fight_1 = Arena.fight(list_knights[0], list_knights[2])
+    fight_2 = Arena.fight(list_knights[1], list_knights[3])
```
ArsenPidhoretskyi, line 16: It is not good to get by id because I can not say that always knight will be in same order. Use dict instead

app/preparation.py
```diff
@@ -0,0 +1,24 @@
...
+
+    def use_potion(self, potion: dict) -> None:
+        if potion is not None:
+            for key, value in potion["effect"].items():
+                if key == "power":
+                    self.power += value
+                if key == "protection":
+                    self.hp += value
+                if key == "hp":
+                    self.hp += value
```
ArsenPidhoretskyi, line 24: You can use getattr and setattr
No need to duplicate)

ArsenPidhoretskyi (changes requested):

app/preparation.py
AlexandrZhezheria: I don’t understand how to write, help please

app/battle_preparation.py
```diff
@@ -0,0 +1,11 @@
+from app.knight_class import Knight
+
+
+def battle_preparation(knight: Knight, knight_dict: dict) -> object:
+    knight.protection += sum([_["protection"] for _ in knight_dict["armour"]])
```
Grayder0152, line 5: Use `_` name only when it is not necessary, pls rename

Grayder0152 (changes requested):

app/battle_preparation.py
```diff
@@ -0,0 +1,11 @@
+from app.knight_class import Knight
+
+
+def battle_preparation(knight: Knight, knight_dict: dict) -> object:
+    knight.protection += sum([armor["protection"] for armor in knight_dict["armour"]])
```
vasylhnatiuk, line 5: you can use `knight.protection += sum(armor["protection"] for armor in knight_dict["armour"]) ` . because it is useless to create list here

vasylhnatiuk (changes requested):

Grayder0152 (approved):

PR #237, opened 2023-03-15
'Solution'

HalynaPetrova:


app/battle_preparation.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def battle_preparation(knight: Knight, knight_dict: dict) -> object:
+    knight.protection += sum([_["protection"] for _ in knight_dict["armour"]])
+    knight.power += knight_dict["weapon"]["power"]
+    if knight_dict["potion"]:
+        knight.power += knight_dict["potion"]["effect"]["power"]
+        knight.hp += knight_dict["potion"]["effect"]["hp"]
+        knight.protection += \
+            knight_dict["potion"]["effect"].get("protection", 0)
```
ArsenPidhoretskyi, line 11: Use loop instead of duplicate code
And setattr

app/fight.py
```diff
@@ -0,0 +1,7 @@
+from app.knight_class import Knight
+
+
+def fight(knight: Knight, enemy: Knight) -> None:
+    knight.hp -= enemy.power - knight.protection
+    knight.hp = 0 if knight.hp <= 0 else knight.hp
+    enemy.hp = 0 if enemy.hp <= 0 else enemy.hp
```
ArsenPidhoretskyi, line 7: How enemy hp can be less <= 0?

ArsenPidhoretskyi (changes requested):

PR #241, opened 2023-03-19
solution

nickkozlov90:


app/knight.py
```diff
@@ -0,0 +1,43 @@
...
+class Knight:
+
+    def __init__(
+            self,
+            name: str,
+            power: int,
+            hp: int,
+            armour: list,
+            weapon: dict,
+            potion: Union[dict, None]
```
Bohdan-Salamakha, line 14: line 12
add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contain string elements has typehint list[str]

line 14
use typehint like: Optional[YOUR_VALUE_TYPE] = None
for example: list that contain string elements and can be None has typehint Optional[list[str]]

olena-hul:
Good job!

PR #247, opened 2023-03-25
Solution

VladyslavKaminskyi:


app/items.py
```diff
@@ -0,0 +1,18 @@
...
+        self.part = armour_conf["part"]
+        self.protection = armour_conf["protection"]
+
+
+class Potion:
+    def __init__(self, potion_conf: dict) -> None:
+        self.name = potion_conf["name"]
+        self.protection_effect = potion_conf["effect"].get("protection")
+        self.power_effect = potion_conf["effect"].get("power")
+        self.hp_effect = potion_conf["effect"].get("hp")
```
ArsenPidhoretskyi, line 12: Useless class. Code is so hard due to this. Because If we will have 100 effects???

app/knight.py
```diff
@@ -0,0 +1,73 @@
...
+        self.protection = 0
+        if not knight_config["armour"]:
+            self.armours = None
+        else:
+            self.armours = [Armour(item) for item in knight_config["armour"]]
+        if knight_config["potion"] is None:
+            self.potion = None
+        else:
+            self.potion = Potion(knight_config["potion"])
+        self.weapon = Weapon(knight_config["weapon"])
```
ArsenPidhoretskyi, line 19: Just in init apply all armor, weapon, potion. No need to save it inside Knight

ArsenPidhoretskyi (changes requested):

PR #256, opened 2023-03-25
KNIGHTS FIGHTING

volodinsy:


app/main.py
```diff
@@ -1,214 +1,26 @@
...
+    knights["mordred"]["hp"] -= (knights["lancelot"]["power"]
+                                 - knights["mordred"]["protection"])
     # 2 Arthur vs Red Knight:
-    arthur["hp"] -= red_knight["power"] - arthur["protection"]
-    red_knight["hp"] -= arthur["power"] - red_knight["protection"]
-
+    knights["arthur"]["hp"] -= (knights["red_knight"]["power"]
+                                - knights["arthur"]["protection"])
+    knights["red_knight"]["hp"] -= (knights["arthur"]["power"]
+                                    - knights["red_knight"]["protection"])
```
katevasilenko, line 16: Does that make sense to create another function that takes two knights as an arguments and here in battle call it and pass two different pair of knights (maybe if loop). In that case you’ll completely avoid self repeating

app/preparation.py
```diff
@@ -0,0 +1,30 @@
+from typing import Callable
+
+
+def preparation(func: Callable) -> Callable:
+    def inner(knights: None) -> None:
```
katevasilenko, line 5: Fix type hints please

katevasilenko (changes requested):

app/main.py
```diff
@@ -1,214 +1,26 @@
...
-        lancelot["hp"] = 0
-
-    if mordred["hp"] <= 0:
-        mordred["hp"] = 0
-
+    knights["lancelot"]["hp"] -= (knights["mordred"]["power"]
+                                  - knights["lancelot"]["protection"])
+    knights["mordred"]["hp"] -= (knights["lancelot"]["power"]
+                                 - knights["mordred"]["protection"])
     # 2 Arthur vs Red Knight:
```
katevasilenko, line 12: Also you don’t need comments anymore in your final version

katevasilenko (changes requested):

PR #258, opened 2023-03-30
Added better solution

antonpiddubnyi:


app/Knights.py
```diff
@@ -0,0 +1,38 @@
...
+        # add potion
+        if self.potion is not None:
+            if "power" in self.potion["effect"]:
+                self.power += self.potion["effect"]["power"]
+
+            if "protection" in self.potion["effect"]:
+                self.protection += self.potion["effect"]["protection"]
+
+            if "hp" in self.potion["effect"]:
+                self.hp += self.potion["effect"]["hp"]
```
ArsenPidhoretskyi, line 38: Instead of duplicate code use loop with getattr and setattr

app/main.py
```diff
@@ -1,214 +1,25 @@
...
+                knight["hp"],
+                knight["armour"],
+                knight["weapon"],
+                knight["potion"])
+        for knight in knights.values()
+    ]
+    for knight in competitors:
+        knight.add_equipment()
+    Battle.competition(competitors[0], competitors[2])
+    Battle.competition(competitors[1], competitors[3])
```
ArsenPidhoretskyi, line 18: Not good to do this by indexes. If in config will be another order? Use dict comprehension instead of list

ArsenPidhoretskyi (changes requested):

Grayder0152 (approved):

PR #264, opened 2023-04-07
the best solution

Munkoor:


app/knights_preparation.py
```diff
@@ -0,0 +1,22 @@
...
+    knight["protection"] = 0
+    for new_value in knight["armour"]:
+        knight["protection"] += new_value["protection"]
+
+    # apply weapon
+    knight["power"] += knight["weapon"]["power"]
+
+    # apply potion if exist
+    if knight["potion"] is not None:
+        if "power" in knight["potion"]["effect"]:
```
vasylhnatiuk, line 12: use for loop with knight["potion"]["effect"] instead of repeating your code

vasylhnatiuk (changes requested):

app/knights_preparation.py
Munkoor: Як використати цикл тут?

vasylhnatiuk: if knight["potion"]:
    affected_attributes = ["power", "protection", "hp"]
    for attribute in affected_attributes:
        if attribute in knight["potion"]["effect"]:
            knight[attribute] += knight["potion"]["effect"][attribute] like that

vasylhnatiuk (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #265, opened 2023-04-09
Solution

jinjo0222988:


app/characters/knight.py
```diff
@@ -0,0 +1,23 @@
...
+        self.protection = 0
+
+    def prepare_for_battle(self) -> None:
+        for armor_element in self.armour:
+            self.protection += armor_element["protection"]
+
+        self.power += self.weapon["power"]
+
+        if self.potion is not None:
+            if "power" in self.potion["effect"]:
```
vasylhnatiuk, line 18: use for loop with self.potion["effect"] here

vasylhnatiuk (changes requested):

app/characters/knight.py
```diff
@@ -0,0 +1,24 @@
...
+        self.power += self.weapon["power"]
+
+        if self.potion is not None:
+            for key in self.potion["effect"]:
+                if key == "power":
+                    self.power += self.potion["effect"][key]
+                if key == "protection":
+                    self.protection += self.potion["effect"][key]
+                if key == "hp":
+                    self.hp += self.potion["effect"][key]
```
Grayder0152, line 24: or you can use `setattr` function to avoid those ifs

Grayder0152 (approved):

app/characters/knight.py
jinjo0222988: I don't understand, why "setattr" better than "if" ?  please explain.

PR #267, opened 2023-04-09
Develop

andy77andy:


app/knights_stats1.py
```diff
@@ -0,0 +1,36 @@
...
+        self.name = name
+        self.hp = hp
+        self.power = power
+        self.protection = protection
+        TheKnight.list_of_knights.append(self)
+
+    @staticmethod
+    def count_stats(stats: dict) -> list:
+        contenders = []
+        for knight in stats.values():
```
Vasyl-Smutok, line 22: better to use `get`

Please fix here and in other places

Vasyl-Smutok (changes requested):
also 
`knights_stats1.py` bad name
did you use `1` at the end

Vasyl-Smutok (approved):
Good Job 👍

PR #268, opened 2023-04-10
Solution_v1

bohdan-yatsyna:


app/knights/unprepared_knight.py
```diff
@@ -0,0 +1,42 @@
...
+            )
+            print(f"{self.name} equipped his armour")
+        else:
+            print(f"{self.name} will fight without armour")
+
+        self.power += self.weapon.get("power", 0)
+        print(f"{self.name} prepared his weapon")
+
+        if self.potion:
+            if "power" in self.potion["effect"]:
```
vasylhnatiuk, line 27: use for loop with self.potion["effect"] and settatr here

vasylhnatiuk (changes requested):

app/knights/unprepared_knight.py
bohdan-yatsyna: Noted.

ArsenPidhoretskyi (approved):
Well done!

PR #270, opened 2023-04-11
Solution

Ida83:


app/knight.py
```diff
@@ -0,0 +1,24 @@
+class Knight:
+    def __init__(self, name: str, power: int, hp: int) -> None:
+        self.name = name
+        self.power = power
+        self.hp = hp
+
+    def using_armour(self, armour: list) -> int:
+        for part in armour:
```
Bohdan-Salamakha, line 8: It's better to use a more specific type hint for armour. Instead of using list, use List[Dict[str, int]] from typing module to provide a more detailed type hint.

app/knight.py
```diff
@@ -0,0 +1,24 @@
...
+            self.hp += part["protection"]
+        return self.hp
+
+    def using_weapon(self, weapon_power: int) -> int:
+        self.power += weapon_power
+        return self.power
+
+    def using_potion(self, potion: dict) -> callable:
```
Bohdan-Salamakha, line 16: For the return type hint, use Tuple[int, int] instead of callable. Import Tuple from the typing module.

Bohdan-Salamakha (changes requested):
v1

app/knight.py
Ida83: can I do annotation without importing form typing module?

app/knight.py
EdAlekseiev: If you use Python 3.10+ you can use list, dict in this case.

EdAlekseiev (approved):
Good job

PR #273, opened 2023-04-13
Develop

PetroHavrysh:


vasylhnatiuk (changes requested):
battle must always accept knightsConfig, and then you must perform all actions in it based on knightsConfig. And these inputs can change, that is, rewrite the code

PR #274, opened 2023-04-15
SOLUTION

VladyslavKaminskyi:


app/knight.py
```diff
@@ -0,0 +1,36 @@
+from __future__ import annotations
+
+
+class Knight:
+    def __init__(self, ser: dict) -> None:
```
vasylhnatiuk, line 5: ser? please rename this variable

vasylhnatiuk (changes requested):

PR #281, opened 2023-04-26
Solution

OlehOryshchuk:
I changed "arthur": {"name": "Arthur", to "arthur": {"name": "Artur", because in test package it is Artur

### DIFF
--- a/app/battle/battle.py
+++ b/app/battle/battle.py
@@ -1,6 +1,6 @@
 class Battle:
     battles_result = {"Lancelot": 0,
-                      "Artur": 0,
+                      "Arthur": 0,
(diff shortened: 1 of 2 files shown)


app/main.py
```diff
@@ -1,3 +1,6 @@
+from app.battle.battle import Battle
+from app.prepare_to_battle.knight_is_ready import IsReady
+
 KNIGHTS = {
```
EdAlekseiev, line 4: Move this variable to a separate file, please.

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #291, opened 2023-04-30
Solution

VladyslavKaminskyi:


app/knight.py
```diff
@@ -0,0 +1,36 @@
+from __future__ import annotations
+
+
+class Knight:
+    def __init__(self, ser: dict) -> None:
+        self.name = ser["name"]
+        self.power = ser["power"] + ser["weapon"].get("power")
+        self.hp = ser["hp"]
+        self.protection = sum([armour["protection"]
+                               for armour in ser["armour"]])
```
Vasyl-Smutok, line 10: no need to create 
```suggestion
        self.protection = sum(armour["protection"]
                               for armour in ser["armour"])
```

Vasyl-Smutok (changes requested):
also use `get`  everywhere

PR #297, opened 2023-05-13
Implement knights fighting task

vitalii-babiienko:


app/knight.py
```diff
@@ -0,0 +1,33 @@
...
+        self.protection = sum(
+            armour_part["protection"] for armour_part in knight["armour"]
+        )
+
+        if knight["potion"] is not None:
+            potion_effects = knight["potion"].get("effect")
+            self.power += potion_effects.get("power", 0)
+            self.hp += potion_effects.get("hp", 0)
+            self.protection += potion_effects.get("protection", 0)
+
```
ArsenPidhoretskyi, line 18: Use for loop with getattr and setattr instead of duplicate code

ArsenPidhoretskyi (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #299, opened 2023-05-14
'Solution'

ilizh:


app/knights/knight.py
```diff
@@ -0,0 +1,25 @@
+from typing import Union
+
+
+class Knights:
+    def __init__(self, name: str, power: int, hp: int, weapon: str,
+                 armor: list,
```
Bohdan-Salamakha, line 6: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contains string elements has typehint list[str]

Bohdan-Salamakha (approved):

PR #300, opened 2023-05-14
'Solution'

Vadim345345:


Vasyl-Smutok (changes requested):
try to split code into modules

PR #301, opened 2023-05-15
Solution

IvanMayor:


app/knights_config.py
```diff
@@ -0,0 +1,86 @@
+KNIGHTS = {
+    "lancelot": {
```
olena-hul, line 2: Please resolve conflicts

image

olena-hul (changes requested):

PR #303, opened 2023-05-17
Solution

sulym:


app/knights/prepare_for_battle.py
```diff
@@ -0,0 +1,5 @@
+def prepare_knight_for_battle(*knights) -> None:
+    for knight in knights:
+        knight.get_protection()
```
Bohdan-Salamakha, line 3: you applied protection in the __init__
so this line is redundant

Bohdan-Salamakha (approved):

PR #313, opened 2023-05-24
refectored knights

Oknametra:


app/main.py
```diff
@@ -1,208 +1,28 @@
...
-    if mordred["hp"] <= 0:
-        mordred["hp"] = 0
+    mordred = knights_config["mordred"]
+    Inventory.inventory_application([mordred])
 
-    # 2 Arthur vs Red Knight:
-    arthur["hp"] -= red_knight["power"] - arthur["protection"]
-    red_knight["hp"] -= arthur["power"] - red_knight["protection"]
+    red_knight = knights_config["red_knight"]
+    Inventory.inventory_application([red_knight])
```
vsmutok, line 20: you write the same thing 4 times, maybe you should use loop?

app/battle/battle_logic.py
```diff
@@ -0,0 +1,11 @@
...
+    @staticmethod
+    def battle(knight1: dict, knight2: dict) -> None:
+        knight1["hp"] -= knight2["power"] - knight1["protection"]
+        knight2["hp"] -= knight1["power"] - knight2["protection"]
+
+        if knight1["hp"] <= 0:
+            knight1["hp"] = 0
+
+        if knight2["hp"] <= 0:
+            knight2["hp"] = 0
```
vsmutok, line 11: add return 

```python
    return {
        knight1["name"]: knight1["hp"],
        knight2["name"]: knight2["hp"],
    }
```

vsmutok (changes requested):

app/main.py
```diff
@@ -1,213 +1,22 @@
...
+    Battle.battle(knight_dict["lancelot"], knight_dict["mordred"])
+    Battle.battle(knight_dict["arthur"], knight_dict["red_knight"])
 
     # Return battle results:
     return {
-        lancelot["name"]: lancelot["hp"],
-        arthur["name"]: arthur["hp"],
-        mordred["name"]: mordred["hp"],
-        red_knight["name"]: red_knight["hp"],
+        value.name: value.hp for name, value in knight_dict.items()
```
ArsenPidhoretskyi, line 19: value -> knight

ArsenPidhoretskyi: You can iterate by values also, because name is not used

ArsenPidhoretskyi (changes requested):

app/main.py
Oknametra: understood

PR #319, opened 2023-06-01
Solution

AntonShpakovych:


app/services/battle.py
```diff
@@ -0,0 +1,39 @@
...
+
+        return {knight.name: knight.hp for knight in self.knights}
+
+    @staticmethod
+    def __hit(knight: Knight, power: int) -> Knight:
+        knight.hp -= power - knight.protection
+
+        if knight.hp < 0:
+            knight.hp = 0
+        return knight
```
Oleg-s-s, line 30: You can avoid the overhead of calling a static method for each hit by calculating the damage directly inside the loop.
Also instead of assigning `knight.hp = 0` when it becomes negative, use the `max` function to set it to 0 if it's negative.

app/knights/knight.py
```diff
@@ -0,0 +1,44 @@
...
+        self.armour = Armour.from_list(armour)
+        self.weapon = Weapon(**weapon)
+        self.potion = Potion(**potion) if potion else None
+        self.protection = 0
+        self.prepared_for_fight()
+
+    def prepared_for_fight(self) -> None:
+        self.__dress_armor()
+        self.__wield_weapon()
+        self.__drink_potion()
```
Oleg-s-s, line 32: could u use this method inside init method ?)

Oleg-s-s (changes requested):

app/knights/knight.py
AntonShpakovych: okey

app/main.py
```diff
@@ -1,214 +1,11 @@
...
-    return {
-        lancelot["name"]: lancelot["hp"],
-        arthur["name"]: arthur["hp"],
-        mordred["name"]: mordred["hp"],
-        red_knight["name"]: red_knight["hp"],
-    }
+    return initialized_battle.start()
 
 
 print(battle(KNIGHTS))
```
LiudmylaKulinchenko, line 11: You can remove print

PR #321, opened 2023-06-02
Solution

Andriy-Sydorenko:


app/knights_info/knights.py
```diff
@@ -0,0 +1,63 @@
...
+        opponent.apply_weapon()
+        opponent.apply_potion()
+
+        self.hp -= opponent.power - self.protection
+        opponent.hp -= self.power - opponent.protection
+
+        if self.hp < 0:
+            self.hp = 0
+        if opponent.hp < 0:
+            opponent.hp = 0
```
Oleg-s-s, line 63: you can use the `max()` function to set the value to zero if it is negative.

Oleg-s-s (changes requested):

app/knights_info/knights.py
```diff
@@ -0,0 +1,61 @@
...
+                 weapon: dict,
+                 potion: dict = None) -> None:
+        self.name = name
+        self.power = power
+        self.hp = hp
+        self.protection = 0
+        self.armour = [Armour(armour_info["part"],
+                              armour_info["protection"])
+                       for armour_info in armour]
+        self.weapon = Weapon(weapon.get("name"), weapon.get("power"))
```
LiudmylaKulinchenko, line 24: U can just unpack the dict `weapon` here. You anyway will have an error during creting a Weapon instance if, for example, `weapon.get("name")` will give a `None`. So you`re not predicting an error with `get` method. Thats why you can just unpack the dict

app/knights_info/knights.py
```diff
@@ -0,0 +1,61 @@
...
+        if potion is not None:
+            self.potion = Potion(potion.get("name"), potion.get("effect"))
```
LiudmylaKulinchenko, line 26: U can just unpack the dict also here

app/knights_info/knights.py
```diff
@@ -0,0 +1,61 @@
...
+        else:
+            self.potion = potion
+
+    def apply_armour(self) -> None:
+        protection = 0
+        if self.armour:
+            protection = sum([part.protection for part in self.armour])
```
LiudmylaKulinchenko, line 33: `sum` can work with generators, so [] are not needed

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,214 +1,28 @@
...
-    # apply armour
-    mordred["protection"] = 0
-    for a in mordred["armour"]:
-        mordred["protection"] += a["protection"]
 
-    # apply weapon
-    mordred["power"] += mordred["weapon"]["power"]
+def battle(knights_config: dict) -> dict:
+    knights = {}
+    for knight_name, knight_info in knights_config.items():
```
LiudmylaKulinchenko, line 6: Create a new function for that

app/main.py
```diff
@@ -1,214 +1,28 @@
...
-
-    # check if someone fell in battle
-    if arthur["hp"] <= 0:
-        arthur["hp"] = 0
-
-    if red_knight["hp"] <= 0:
-        red_knight["hp"] = 0
-
-    # Return battle results:
     return {
```
LiudmylaKulinchenko, line 23: Try to rewrite in one line

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done! 🚀🚀

PR #330, opened 2023-06-15
'Solution'

KirillMelanich:


app/main.py
```diff
@@ -1,214 +1,40 @@
...
+    knights = []
+    for attributes in attributes.values():
+        knights.append(Knight(name=attributes["name"],
+                              power=attributes["power"],
+                              hp=attributes["hp"],
+                              armour=attributes["armour"],
+                              weapon=attributes["weapon"],
+                              potion=attributes["potion"]))
+    for knight in knights:
+        knight.preparation()
```
Grayder0152, line 14: do it in the first loop

Grayder0152 (changes requested):
Pls, resolve comment and conflicts

app/main.py
```diff
@@ -130,85 +133,42 @@ def battle(knightsConfig):
...
-    # check if someone fell in battle
-    if lancelot["hp"] <= 0:
-        lancelot["hp"] = 0
 
-    if mordred["hp"] <= 0:
-        mordred["hp"] = 0
+def battle(knights: dict) -> dict:
+    knights = create_knights(knights)
+    fight(knights[0], knights[2])
+    fight(knights[1], knights[3])
```
ArsenPidhoretskyi, line 167: Due to task fighting must be based on name, not indexes. If I change order of elements i will get wrong output

ArsenPidhoretskyi (approved):
Well done!

PR #343, opened 2023-06-28
Solution

Anastasia-Su:


app/battle_prepare/battle_prepare.py
```diff
@@ -0,0 +1,23 @@
+from app.constants.property_list import PROPERTY_LIST
+
+
+def battle_prepare(knight: dict, config: dict) -> None:
+    def add_potion_effect(prop: str) -> None:
+        potion_effect = knight["potion"]["effect"]
+        if prop in potion_effect:
+            knight[prop] += potion_effect[prop]
+
+    knight = config[knight]
```
vsmutok, line 10: why you need it? 

you already have a knight whose function is accepted

app/battle_prepare/battle_prepare.py
```diff
@@ -0,0 +1,23 @@
+from app.constants.property_list import PROPERTY_LIST
+
+
+def battle_prepare(knight: dict, config: dict) -> None:
+    def add_potion_effect(prop: str) -> None:
+        potion_effect = knight["potion"]["effect"]
```
vsmutok, line 6: I think better to user `get`

app/constants/duel_list.py
```diff
@@ -0,0 +1 @@
+DUEL_PAIRS = [("lancelot", "mordred"), ("arthur", "red_knight")]
```
vsmutok, line 1: I think you can combine `app/constants/property_list.py` with this module

app/constants/knights_dict.py
```diff
@@ -0,0 +1,86 @@
...
+                "protection": 15,
+            },
+            {
+                "part": "breastplate",
+                "protection": 20,
+            },
+            {
+                "part": "boots",
+                "protection": 10,
+            }
```
vsmutok, line 29: put data to json file

vsmutok (changes requested):

app/battle_prepare/battle_prepare.py
```diff
@@ -0,0 +1,21 @@
+from app.constants.lists import PROPERTY_LIST
+
+
+def battle_prepare(knight: dict) -> None:
+    def add_potion_effect(prop: str) -> None:
+        potion_effect = knight.get("potion", {}).get("effect")
```
ArsenPidhoretskyi, line 6: It is always true due to:
`if "potion" in knight and knight["potion"] is not None:`

app/main.py
```diff
@@ -1,214 +1,22 @@
...
 
-    # 2 Arthur vs Red Knight:
-    arthur["hp"] -= red_knight["power"] - arthur["protection"]
-    red_knight["hp"] -= arthur["power"] - red_knight["protection"]
 
-    # check if someone fell in battle
-    if arthur["hp"] <= 0:
-        arthur["hp"] = 0
+def battle(config: dict) -> dict:
+    for knight in config:
```
ArsenPidhoretskyi, line 8: Use .values

ArsenPidhoretskyi (changes requested):

app/battle_prepare/battle_prepare.py
Anastasia-Su: Do you mean that "if prop in potion_effect:" can be removed? But it is impossible for this case, since not all the effects from PROPERTY_LIST are present in KNIGHTS data. red_knight["potion"] does not have "protection" effect.

PR #345, opened 2023-06-29
Solution

ShevchukVitalii:


app/knight.py
```diff
@@ -0,0 +1,33 @@
...
+        self.weapon = weapon
+        self.protection = sum(element["protection"] for element in armour)
+        if potion is not None:
+            self.apply_potion_effects(potion)
+
+    def apply_potion_effects(self, potion: dict) -> None:
+        effects = potion.get("effect", {})
+        self.power += effects.get("power", 0)
+        self.hp += effects.get("hp", 0)
+        self.protection += effects.get("protection", 0)
```
ArsenPidhoretskyi, line 24: Use forloop and setattr and getattr

app/knight.py
```diff
@@ -0,0 +1,33 @@
...
+
+    @staticmethod
+    def fight(knight1: Knight, knight2: Knight) -> None:
+        if knight1.protection < knight2.power:
```
ArsenPidhoretskyi, line 28: No need this if. Remove it
Also with line 31

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (changes requested):
Changes requested

app/knight.py
ShevchukVitalii: Done

ArsenPidhoretskyi (approved):
Well done!

PR #355, opened 2023-07-07
solution

oleksiikolii:


app/knights/items/armour.py
```diff
@@ -0,0 +1,15 @@
...
+    part: str
+    protection: int
+
+    @classmethod
+    def create(cls, dict_: dict) -> list | None:
+        if dict_["armour"]:
+            res = []
+            for piece in dict_["armour"]:
+                res.append(cls(piece["part"], piece["protection"]))
+            return res
```
Kirontiko, line 15: u can transform it into generator:
return [... for ... in ...]

app/knights/knight.py
```diff
@@ -0,0 +1,59 @@
...
+            weapon=Weapon.create(knight),
+            potion=Potion.create(knight)
+        )
+
+    def __apply_stats(self) -> None:
+        self.power += self.weapon.power
+
+        if self.armour:
+            for armour in self.armour:
+                self.protection += armour.protection
```
Kirontiko, line 49: Same

Kirontiko (approved):

app/knights/knight.py
```diff
@@ -0,0 +1,58 @@
...
+
+    @property
+    def hp(self) -> int:
+        return self._hp
+
+    @hp.setter
+    def hp(self, value: int) -> None:
+        if value <= 0:
+            self._hp = 0
+            return
```
YuriiKindrat, line 26: it seems to me that it's better to use else here instead of blank return

YuriiKindrat (changes requested):

YuriiKindrat (changes requested):
also, all tests should pass )

Oleg-s-s (approved):
:+1:

PR #364, opened 2023-07-20
Solution

sachavskyi:


app/main.py
```diff
@@ -1,213 +1,61 @@
...
-        arthur["hp"] = 0
-
-    if red_knight["hp"] <= 0:
-        red_knight["hp"] = 0
+from app.knight import Knight
+from app.config import KNIGHTS
+
+
+def battle(knights: dict) -> dict:
+    knight1 = Knight(
```
Polyakiv-Andrey, line 6: when creating instances of knights, a certain pattern is observed that can be expressed through a list generator, and then, through multiple assignments, pass each knight to a variable

Polyakiv-Andrey (changes requested):

app/knight.py
```diff
@@ -0,0 +1,40 @@
...
+        if self.armour:
+            for equipment in self.armour:
+                self.protection += equipment.get("protection")
+
+    def set_weapon(self) -> None:
+        self.power += self.weapon.get("power")
+
+    def set_potion(self) -> None:
+        if self.potion:
+            if "power" in self.potion.get("effect"):
```
Dimosphen1, line 30: You do not need to call `self.potion.get("effect")` three times throughout these if conditions, you can call it once and write it to a separate variable.

app/knight.py
```diff
@@ -0,0 +1,40 @@
...
+                self.power += self.potion.get("effect").get("power")
+            if "protection" in self.potion.get("effect"):
+                self.protection += self.potion.get("effect").get("protection")
+            if "hp" in self.potion.get("effect"):
+                self.hp += self.potion.get("effect").get("hp")
```
Dimosphen1, line 35: Regarding these three if conditions, avoid using them in the following way:
- Create a list of `effects` where you will have names of all effects;
- Iterate through this list;
- Use the name of the effect together with `getattr` and `setattr` to update the values.

It will allow to make code more Pythonic and avoid using code duplication.

app/knight.py
```diff
@@ -0,0 +1,40 @@
+class Knight:
+    protection = 0
```
Dimosphen1, line 2: There is no need to have `protection` here as a class attribute because it rather relates to an instance, place it in the `__init__`. Besides that, you are calling `protection` via `self` in code, which is why `__init__` is a better place to have this attribute.

app/knight.py
```diff
@@ -0,0 +1,40 @@
...
+    def set_potion(self) -> None:
+        if self.potion:
+            if "power" in self.potion.get("effect"):
+                self.power += self.potion.get("effect").get("power")
+            if "protection" in self.potion.get("effect"):
+                self.protection += self.potion.get("effect").get("protection")
+            if "hp" in self.potion.get("effect"):
+                self.hp += self.potion.get("effect").get("hp")
+
+    def preparation(self) -> None:
```
Dimosphen1, line 37: Method names should represent some action, which is why using gerund `preparation` in naming is not the best approach. Rename it to the verb `prepare`.

app/main.py
```diff
@@ -1,213 +1,33 @@
...
-        red_knight["hp"] = 0
+from app.knight import Knight
+from app.config import KNIGHTS
+
+
+def battle(knights: dict) -> dict:
+    knight1 = Knight(*[value for value in knights.get("lancelot").values()])
+    knight2 = Knight(*[value for value in knights.get("arthur").values()])
+    knight3 = Knight(*[value for value in knights.get("mordred").values()])
+    knight4 = Knight(*[value for value in knights.get("red_knight").values()])
```
Dimosphen1, line 9: Shorten this code to `Knight(*knights.get("lancelot").values())`, there is no need for list comprehension here.

app/main.py
```diff
@@ -1,213 +1,33 @@
...
+
+    def fight(first_knight: Knight, second_knight: Knight) -> None:
+        first_knight.hp -= second_knight.power - first_knight.protection
+        second_knight.hp -= first_knight.power - second_knight.protection
+        if first_knight.hp <= 0:
+            first_knight.hp = 0
+        if second_knight.hp <= 0:
+            second_knight.hp = 0
```
Dimosphen1, line 17: You could place this logic in a method in the `Knight` class.

Dimosphen1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #370, opened 2023-07-26
Solution

EvgenMilokumov:


app/fighters/knight.py
```diff
@@ -0,0 +1,37 @@
...
+        potion: dict,
+    ) -> object:
+        self.name = name
+        self.power = power
+        self.hp = hp
+        self.protection = 0
+        self.armour = armour
+        self.weapon = weapon
+        self.potion = potion
+        self.dict_of_all_knights_instances[self.name] = self
```
ArsenPidhoretskyi, line 20: Why you need to save it?

app/fighters/knight.py
```diff
@@ -0,0 +1,37 @@
...
+                knight["name"],
+                knight["power"],
+                knight["hp"],
+                knight["armour"],
+                knight["weapon"],
+                knight["potion"],
+            )
+            dict_of_knights_instances[knight["name"]] = knight.get(
+                knight["name"], new_knight
+            )
```
ArsenPidhoretskyi, line 36:
```suggestion
            dict_of_knights_instances[knight["name"]] = new_knight
```

app/fighters/perks.py
```diff
@@ -0,0 +1,44 @@
+from app.fighters.knight import Knight
+
+
+class Perk:
+    @staticmethod
+    def apply_armour(knight: Knight) -> None:
+        armour_effect = 0
+        for part in knight.armour:
+            armour_effect += part["protection"]
```
ArsenPidhoretskyi, line 9: Use sum function

app/fighters/perks.py
```diff
@@ -0,0 +1,44 @@
...
+            for effect, value in potion_effects.items():
+                if effect == "power":
+                    knight.power += value
+                    potion_effects_result += f"+ power: {value}, "
+                if effect == "hp":
+                    knight.hp += value
+                    potion_effects_result += f"+ hp: {value}, "
+                if effect == "protection":
+                    knight.protection += value
+                    potion_effects_result += f"+ protection: {value} "
```
ArsenPidhoretskyi, line 35: Use forloop instead of duplicate code

app/main.py
```diff
@@ -1,214 +1,20 @@
...
-    # apply armour
-    mordred["protection"] = 0
-    for a in mordred["armour"]:
-        mordred["protection"] += a["protection"]
 
-    # apply weapon
-    mordred["power"] += mordred["weapon"]["power"]
+def battle(base_knights_config: dict) -> dict:
+    fighters = Knight.create_dict_of_knights_instances(base_knights_config)
+    lancelot, arthur, mordred, red_knight = fighters.values()
```
ArsenPidhoretskyi, line 7: Better to get by names because if input order will be another the result will be incorrect

ArsenPidhoretskyi (changes requested):

app/fighters/knight.py
EvgenMilokumov: thanks

app/fighters/knight.py
EvgenMilokumov: I'm creating these attributes and storing them with object in dict of all object of the class Knight in order to have possibiity use them when applying perks receiving in perks methods only knight object and also if somebody wants to create the knight object he can do this with all attributes but not by sending, something like this 

otherwise, I guess, my perks methods should receive as an argument base_knights_config and use apply perks values to hp and power, but I've chose maybe bad way(

app/main.py
EvgenMilokumov: done

app/fighters/perks.py
EvgenMilokumov: done

app/fighters/perks.py
EvgenMilokumov: done

app/fighters/knight.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def __init__(
+        self,
+        name: str,
+        power: int,
+        hp: int,
+        armour: list[dict],
+        weapon: dict,
+        potion: dict,
+    ) -> object:
```
vsmutok, line 12:
```suggestion
    ) -> None:
```

app/fighters/knight.py
```diff
@@ -0,0 +1,35 @@
...
+        dict_of_knights_instances = {}
+        for knight in knights.values():
+            new_knight = Knight(
+                knight["name"],
+                knight["power"],
+                knight["hp"],
+                knight["armour"],
+                knight["weapon"],
+                knight["potion"],
+            )
```
vsmutok, line 33:
```suggestion
            new_knight = Knight(**knight)
```

vsmutok (changes requested):

app/fighters/knight.py
EvgenMilokumov: thanks

app/fighters/knight.py
EvgenMilokumov: done

vsmutok (approved):
good job 👍

PR #371, opened 2023-07-29
Solution

Starchild666:


app/knights.py
```diff
@@ -0,0 +1,56 @@
...
+        return protection
+
+    @staticmethod
+    def prepare_weapon(dict_knight: dict) -> int:
+        return dict_knight["weapon"]["power"]
+
+    def potion_shot(self, dict_knight: dict) -> None:
+        potion = dict_knight.get("potion")
+        if potion:
+            self.power += potion["effect"].get("power", 0)
```
vasylhnatiuk, line 32: try to use loop and `setattr` here instead of duplicating code

vasylhnatiuk (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #372, opened 2023-07-30
Solution

AlexPelevin99:


app/battle/battle.py
```diff
@@ -0,0 +1,22 @@
...
+    def __init__(self, knight_1: Knight, knight_2: Knight) -> None:
+        self.knight_1 = knight_1
+        self.knight_2 = knight_2
+
+    def fight(self) -> dict:
+        self.knight_1.hp -= self.knight_2.power - self.knight_1.protection
+        self.knight_2.hp -= self.knight_1.power - self.knight_2.protection
+
+        if self.knight_1.hp <= 0:
+            self.knight_1.hp = 0
```
pavlejviki, line 14: You can use max(self.knight_1.hp, 0) in this case.

app/knight/knight.py
```diff
@@ -0,0 +1,39 @@
...
+        self.armours = []
+        self.potion = None
+
+        Knight.knights.append(self)
+
+    def add_armour(self, armour: Armour) -> Knight:
+        if armour not in self.armours:
+            self.armours.append(armour)
+            self.protection += armour.protection
+        return self
```
pavlejviki, line 32: Why does this method return self?

app/knight/knight.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def set_potion(self, potion: Potion) -> Knight:
+        self.power += potion.effect.power
+        self.protection += potion.effect.protection
+        self.hp += potion.effect.hp
+
+        return self
```
pavlejviki, line 39: Why does this method return self? Do you really need it?

app/knight/knight.py
```diff
@@ -0,0 +1,39 @@
+from __future__ import annotations
+from app.armour.armour import Armour
+from app.weapon.weapon import Weapon
+from app.potion.potion import Potion
+
+
+class Knight:
+
+    knights: list[Knight] = []
```
pavlejviki, line 9: Why did you create this class attribute?

app/main.py
```diff
@@ -1,214 +1,55 @@
...
+        )
+
+        knight.set_potion(potion)
+
+    return knight
+
+
+def battle(knights_config: dict) -> dict:
+    knights = {}
+    for key, knight_config in knights_config.items():
```
pavlejviki, line 43: key is not very descriptive variable name in this case.

app/main.py
```diff
@@ -1,214 +1,55 @@
...
+    battle_results = {}
+    for i in range(0, 2):
+        if i == 0:
+            battle_obj = Battle(knights["lancelot"], knights["mordred"])
+        else:
+            battle_obj = Battle(knights["arthur"], knights["red_knight"])
+
+        battle_results.update(battle_obj.fight())
+
+    return battle_results
```
pavlejviki, line 55:
```suggestion
   knight_pairs = [("lancelot", "mordred"), ("arthur", "red_knight")]
   battle_results = {}
   
    for knight1, knight2 in knight_pairs:
        battle_obj = Battle(knights[knight1], knights[knight2])
        battle_results.update(battle_obj.fight())

    return battle_results
```

pavlejviki (changes requested):

app/knight/knight.py
AlexPelevin99: it is not required, but why not ?

app/knight/knight.py
AlexPelevin99: Returning self from a method simply means that method returns a reference to the instance object on which it was called

Polyakiv-Andrey (approved):
Good job!

PR #376, opened 2023-08-08
Implemented solution

RPodolets:


eduardhabryd (approved):
GJ!

app/equipment/potion.py
```diff
@@ -0,0 +1,12 @@
+class Potion:
+    class Effect:
```
Kroll410, line 2: Separate class from here

app/knight.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def add_potion(self, potion: dict) -> None:
+        if potion:
+            name, effect = potion.values()
+            self.potion = Potion(name, effect)
+
+    def drink_potion(self) -> None:
+        if self.potion:
+            for effect in self.potion.effects:
+                match effect.attr:
```
Kroll410, line 43: match case is good for dynamic unpacking compare operations, there you're trying to apply it like a c-style switch case. Use if-else blocks here, you will remove redundant 1 indent and increase readability

Kroll410 (changes requested):

Kroll410 (approved):

PR #378, opened 2023-08-09
Nikita Karpenko py-23-july

ORENYT:
Start point

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -211,4 +211,5 @@ def battle(knightsConfig):
     }
 
 
-print(battle(KNIGHTS))
+if __name__ == "__main__":
+    print(battle(KNIGHTS))


app/ammunition/potion.py
```diff
@@ -0,0 +1,25 @@
...
+    Knight's potion.
+
+    Something can be not mentioned, so checked for None.
+    """
+    def __init__(self, name: str, effect: dict) -> None:
+        self.name = name
+        self.hp_boost = effect.get("hp")
+        self.power_boost = effect.get("power")
+        self.protection_boost = effect.get("protection")
+        if self.hp_boost is None:
```
Kroll410, line 12: You can simply assign variable inline, because get() and pass default parameter to get("hp", 0) for example

app/knights.py
```diff
@@ -0,0 +1,45 @@
+from app.ammunition.armour import Armour
+from app.ammunition.potion import Potion
+from app.ammunition.weapon import Weapon
+
+
+class Knight:
+    """"Knight transformed from incoming dict key to class"""
+    def __init__(self,
```
Kroll410, line 8: Apply next line for each paramter so they look like this https://stackoverflow.com/a/41591509

app/knights_config.py
```diff
@@ -0,0 +1,121 @@
...
+        knight = Knight(
+            name=stats["name"],
+            power=stats["power"],
+            hp=stats["hp"],
+            armour=armour,
+            weapon=Weapon(
+                name=stats["weapon"]["name"],
+                power=stats["weapon"]["power"]
+            ),
+            potion=None if stats.get("potion") is None else Potion(
```
Kroll410, line 115: Use second `get` paramter if potion is None

Kroll410 (changes requested):

app/ammunition/potion.py
ORENYT: Removed none-check. get default variable 0 is setted

app/knights.py
ORENYT: Did this, but now code doesn't look better, as for me.

app/knights_config.py
ORENYT: potion must be None if it is not provided, so default value of None potion is None. Potion class works with None potion correct.

Kroll410: I mean pass the Potion object into second paramter of get method, so you can avoid ternary here

Kroll410 (changes requested):

app/knights_config.py
ORENYT: I don't understand, here is no ternay, here is None if potion is None or it is a Potion object. Passing  stats.get("potion", Potion(args) ) result in error in test. I need to pass None if potion is None.

ORENYT: image

Kroll410 (approved):

PR #379, opened 2023-08-09
solution

innyshka:


ORENYT:
Interesting solution, but why you didn't make more classes? I thought the task is about it

app/logic.py
```diff
@@ -0,0 +1,37 @@
+from __future__ import annotations
```
Kroll410, line 1: This solution works, but you should try to split up logic to functions or classes more widely
image

Kroll410 (changes requested):

app/knight/fight.py
```diff
@@ -0,0 +1,8 @@
+class Battle:
+    @staticmethod
+    def battle(all_kn: list) -> list:
```
Kroll410, line 3: use better namings

app/main.py
```diff
@@ -1,214 +1,23 @@
...
+from app.knight.equipment.armour import Armour
+from app.knight.equipment.weapon import Weapon
+from app.knight.equipment.potion import Potion
+from app.knight.knights import Knights
+from app.knight.fight import Battle
+from app.knight.data import KNIGHTS
+
+
+def battle(config: dict) -> dict:
+    all_knights = [Knights(characteristics["name"],
```
Kroll410, line 10: nextline

app/knight/fight.py
```diff
@@ -0,0 +1,9 @@
+class Battle:
+    @staticmethod
+    def battle(all_knights: list) -> list:
+        for i in range(4):
```
Kroll410, line 4: why range(4)? seems like mock

app/main.py
```diff
@@ -1,214 +1,26 @@
...
+from app.knight.equipment.weapon import Weapon
+from app.knight.equipment.potion import Potion
+from app.knight.knights import Knights
+from app.knight.fight import Battle
+from app.knight.data import KNIGHTS
+
+
+def battle(config: dict) -> dict:
+    all_knights = [
+        Knights(characteristics["name"],
```
Kroll410, line 11: Nextline

Kroll410 (changes requested):
Looks good

Kroll410 (approved):

PR #382, opened 2023-08-10
Oh My God, I did it!!!

maxkatkalov:
Please check


app/Knights.py
```diff
@@ -0,0 +1,90 @@
...
+                    Armour(
+                        fighter["part"],
+                        fighter["protection"]
+                    )
+                )
+
+    def apply_armour(
+        self,
+    ) -> None:
+        if self.armour:
```
Kroll410, line 59: If self.armour is  an empty list, you can omit this condition because loop will be exited automatically

app/battle.py
```diff
@@ -0,0 +1,38 @@
...
+        knights[fighter2].hp -= (
+            knights[fighter1].power - knights[fighter2].protection
+        )
+
+    @staticmethod
+    def is_anyone_alive(
+            knights: Dict[str, Knight]
+    ) -> None:
+        for fighter in knights.values():
+            if fighter.hp <= 0:
```
Kroll410, line 28: Use ternary here to avoid redundant indent

Kroll410 (changes requested):
Good structure

Kroll410 (approved):

PR #383, opened 2023-08-10
Solution

OlegatorLE:


Kroll410 (changes requested):
The logic is OK, but you need to split it more into the functions/classes and call them

app/action.py
```diff
@@ -0,0 +1,14 @@
+def action(knights_dict: dict) -> dict:
```
Kroll410, line 1: This looks non readable

Kroll410: Try to use separate functions that calculates your logic here

Kroll410 (changes requested):
You can use classes for the knights instatiation and all the actions alongside or split the logic so it's easy to re-apply and read

app/action.py
OlegatorLE: I dont want to use classes here. Changed the logic of calculating the results immediately in pairs

Kroll410 (changes requested):
If you want to use functions, use them in the next modules with specific methods created, for ex. 
`main` - get_knight, `preparations` -  `set_protection`, `set_health`, `add_weapon`, `set_potion`, etc. So you should have initialized default knights, populated their fields via the functions and processed any actions after in main.py. The solution natively covers the requirements, but ask you, what if I add more knights in the KNIGTS variable? Give more armor? This will lead to the changes that I proposed above

OlegatorLE:
> If you want to use functions, use them in the next modules with specific methods created, for ex. `preparations` - `set_protection`, `set_health`, `add_weapon`, `set_potion`, etc. So you should have initialized default knights, populated their fields via the functions and processed any actions after in main.py. The solution natively covers the requirements, but ask you, what if I add more knights in the KNIGTS variable? Give more armor? This will lead to the changes that I proposed above

if you add more knights or something else - it will be another task and another solution )

app/main.py
```diff
@@ -1,213 +1,20 @@
...
-    if arthur["hp"] <= 0:
-        arthur["hp"] = 0
 
-    if red_knight["hp"] <= 0:
-        red_knight["hp"] = 0
+def battle(knights: dict) -> dict:
+    # prepare all knights
+    knights_dict = preparations(knights)
+    # knights are fighting here
+    fight(knights_dict["lancelot"], knights_dict["mordred"])
```
Kroll410, line 10: What if I add a new pair of knights? You will be need to add them here as well, so your code still is not reusable

Kroll410 (changes requested):

app/main.py
OlegatorLE: this question is not correct, because in any case, this pair should be written in the code, who will fight with whom. and if you don't even add more knights, but simply change their places, even then you have to rewrite the condition. So make code that can be reused, only with addition input function where you can manually enter the name of the knight and his opponent.

Kroll410: But the knights number are not specific, you have an example in `KNIGHTS` variable. Try to add 2 new knights and rerun your code
image

OlegatorLE: nothing will change because they are just extra knights. They will not affect the solution of the main condition of the task. What do you want me to do neither classes nor functions can solve this problem. because you still need to know who is fighting whom. In condition 1, it fights with 3m. 2 out of 4. If I add 5 and 6 knights, where is the challenge that they have to fight each other instead of 1 out of 5, 2 out of 6? That's why I say - this is not a correct question.

Kroll410 (approved):

PR #384, opened 2023-08-10
'Solution'

hanovskyiv:


app/championship/knights.py
```diff
@@ -0,0 +1,51 @@
...
+        # apply base stat
+        name = knight_stat["name"]
+        hp = knight_stat["hp"]
+        protection = 0
+        power = knight_stat["power"]
+
+        # apply armour
+        if knight_stat["armour"]:
+            protection = sum(
+                [part["protection"] for part in knight_stat["armour"]]
```
Dimosphen1, line 28: You do not need to use list comprehension inside of `sum` here, rewrite code without using it.

app/championship/knights.py
```diff
@@ -0,0 +1,51 @@
...
+            hp: int,
+            protection: int
+    ) -> None:
+        self.name = name
+        self.power = power
+        self.hp = hp
+        self.protection = protection
+
+    @classmethod
+    def add_knights(cls, knight_stat: dict) -> None:
```
Dimosphen1, line 18: You add one knight in this method; rename it to `add_knight` so naming would not be misleading.

app/championship/battles.py
```diff
@@ -0,0 +1,13 @@
...
+class Battle:
+
+    @staticmethod
+    def duel(fighter_1: str, fighter_2: str) -> None:
+        fighters = [Knight.knights[fighter_1], Knight.knights[fighter_2]]
+        fighters[0].hp -= fighters[1].power - fighters[0].protection
+        fighters[1].hp -= fighters[0].power - fighters[1].protection
+        for fighter in fighters:
+            if fighter.hp <= 0:
+                fighter.hp = 0
```
Dimosphen1, line 13: There is no need to arrange a separate class to store one static method. Use a function for this purpose.

app/championship/battle_results.py
```diff
@@ -0,0 +1,2 @@
+def results(members: dict) -> dict:
```
Dimosphen1, line 1: You could place this function and the one from the `Battle` class in a separate file and name it e.g. `battle_functions.py`.

Dimosphen1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #403, opened 2023-08-29
solution

MykolaKrasulin:


Polyakiv-Andrey (changes requested):
flake8 not passed

Dimosphen1 (approved):
LGTM, good job, it is an appropriate implementation using a functional approach. However, in your future project structure consider dividing logic into separate modules to separate functions according to their responsibilities.

PR #414, opened 2023-09-14
'Solution'

nsdNite:


app/data/functions.py
```diff
@@ -0,0 +1,16 @@
...
+        data["name"],
+        data["power"],
+        data["hp"]
+    )
+
+    knight.grab_weapon(data["weapon"])
+    knight.suit_up(data["armour"])
+    knight.drink_potion(data["potion"])
+
+    return knight
```
viktoria-rybenchuk, line 16: It is generally not a good idea to name   a module as functions in Python because it can lead to confusion.

app/data/functions.py
```diff
@@ -0,0 +1,16 @@
+from __future__ import annotations
+from app.data.knight_class import Knight
+
+
+def knight_the_knight(data: dict) -> Knight:
```
viktoria-rybenchuk, line 5: rename this function, it is not clear

viktoria-rybenchuk (changes requested):

app/data/functions.py
nsdNite: renamed
thanks for the comment

app/data/functions.py
nsdNite: done
thanks for comment

Polyakiv-Andrey (approved):
Good job!

PR #421, opened 2023-09-20
Solution

narizhiy:


app/battle_preparations/apply_armour.py
```diff
@@ -0,0 +1,9 @@
+class Armour:
+    @staticmethod
+    def armour_protection(armors: list) -> int:
```
Bohdan-Salamakha, line 3: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contains string elements has typehint list[str]

app/main.py
```diff
@@ -1,214 +1,27 @@
...
+    for index in range(len(knight_list)):
+        if knight_list[index].name == "Lancelot" \
+                or knight_list[index].name == "Arthur":
+            Check.check_knights(
+                Battle.battle(knight_list[index], knight_list[index + 2])
+            )
+    return Check.result_dict
 
 
 print(battle(KNIGHTS))
```
Bohdan-Salamakha, line 27: remove this test print logic

Bohdan-Salamakha (changes requested):

Bohdan-Salamakha (approved):

PR #422, opened 2023-09-28
feat: solution

sashokpilipok7:


app/main.py
```diff
@@ -86,129 +88,21 @@
...
-            red_knight["power"] += red_knight["potion"]["effect"]["power"]
-
-        if "protection" in red_knight["potion"]["effect"]:
-            red_knight["protection"] += red_knight["potion"]["effect"]["protection"]
-
-        if "hp" in red_knight["potion"]["effect"]:
-            red_knight["hp"] += red_knight["potion"]["effect"]["hp"]
-
+    new_knights_dict = {}
+    for key in base_knights_config:
```
vasylhnatiuk, line 94: key is a bad name

app/main.py
```diff
@@ -86,129 +88,21 @@
...
+        person = Knight(base_knights_config[key])
```
vasylhnatiuk, line 95:
```suggestion
        knight = Knight(base_knights_config[key])
```

app/main.py
```diff
@@ -86,129 +88,21 @@
 }
 
 
-def battle(knightsConfig):
+def battle(base_knights_config: dict) -> dict:
```
vasylhnatiuk, line 91:
```suggestion
def battle(knights_config: dict) -> dict:
```

app/knight/main.py
```diff
@@ -0,0 +1,31 @@
...
+        for item in knight_dict["armour"]:
+            self.protection += item["protection"]
+
+        # Add weapons
+        self.power += knight_dict["weapon"]["power"]
+
+        # Add potion
+        if knight_dict["potion"] is not None:
+            effect_values = knight_dict["potion"]["effect"]
+            if "power" in effect_values:
```
vasylhnatiuk, line 19: try to avoid many `if` here, use loop and setattr()

vasylhnatiuk (changes requested):

PR #426, opened 2023-10-01
my implementation

kravetb:
I implemented this task but no one pytest not was completed(

### DIFF
--- a/app/fighters/knights_stat.py
+++ b/app/fighters/knights_stat.py
@@ -1,0 +1,65 @@
+from app.fighters import equipment
+
+
+class Knight:
+    def __init__(
+        self,
+        name: str,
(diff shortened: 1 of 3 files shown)


Bohdan-Salamakha (changes requested):
fix errors in tests

app/fighters/knights_stat.py
YukihiroSM, line 1: I would recommend you to reformat this file using Ctrl + Alt + Shift +L if you are using PyCharm, or RMB -> format file in VSCode. 

There is a couple of formatting mistakes, like separating the line with `\`, or wrong spacing.

YukihiroSM (changes requested):

PR #429, opened 2023-10-04
'Solution'

romanshupak:


Yurnerroo (changes requested):

app/knights/knight.py
```diff
@@ -0,0 +1,40 @@
...
+        self.protection = 0
+        for armour_piece in self.armour:
+            self.protection += armour_piece["protection"]
+
+    def apply_weapon(self) -> None:
+        self.power += self.weapon["power"]
+
+    def apply_potion(self) -> None:
+        if self.potion is not None:
+            if "power" in self.potion["effect"]:
```
vasylhnatiuk, line 35: use loop and setattr() instead of many `if `

vasylhnatiuk (changes requested):

Grayder0152 (approved):

PR #430, opened 2023-10-04
Solution

Naz-iv:


D1st3f (approved):
Looks OK

app/arenas/camelot.py
```diff
@@ -0,0 +1,27 @@
...
+        knight_one.hp -= knight_two.power - knight_one.protection
+        knight_two.hp -= knight_one.power - knight_two.protection
+
+        if knight_one.hp < 0:
+            print(f"{knight_one.name} fall in battle!")
+            knight_one.hp = 0
+
+        if knight_two.hp < 0:
+            print(f"{knight_two.name} fall in battle!")
+            knight_two.hp = 0
```
LiudmylaKulinchenko, line 22: u can use `max(knight_one.hp, 0)`

app/main.py
```diff
@@ -1,214 +1,28 @@
...
+                               .get("Mordred")
+                               .prepare_for_battle())
+    battle_arena.knight_battle(Knight.knights_for_battle
+                               .get("Arthur")
+                               .prepare_for_battle(),
+                               Knight.knights_for_battle
+                               .get("Red Knight")
+                               .prepare_for_battle())
+
+    return battle_arena.battles_results
```
LiudmylaKulinchenko, line 28: create additional variables, it has bad readability now

app/arenas/camelot.py
```diff
@@ -0,0 +1,27 @@
+from app.knights.knight import Knight
+
+
+class Camelot:
+
+    battles_results = {}
```
LiudmylaKulinchenko, line 6: avoid store as a class attribute

LiudmylaKulinchenko (changes requested):

app/arenas/camelot.py
Naz-iv: Thanks you for advice. Seen this construction used many times in morning tasks, but when it's time to use it - I always forget about it.

app/main.py
Naz-iv: I tired to add new variables, was still looking messy. Decided to adjust spacing and looks better to me now. What are your thoughts on it?

app/arenas/camelot.py
Naz-iv: moved storage to main

app/main.py
```diff
@@ -1,214 +1,30 @@
...
+            armour=knight["armour"],
+            weapon=knight["weapon"],
+            potion=knight["potion"])
+        for knight in knights_config.values()]
 
-    # apply weapon
-    lancelot["power"] += lancelot["weapon"]["power"]
+    first_battle_result = battle_arena.knight_battle(
+        Knight.knights_for_battle.get("Lancelot").prep_for_battle(),
+        Knight.knights_for_battle.get("Mordred").prep_for_battle())
```
LiudmylaKulinchenko, line 21: how about create a variable for each knight

app/main.py
```diff
@@ -1,214 +1,30 @@
...
```
LiudmylaKulinchenko, line 20: Also fix formatting here

app/main.py
```diff
@@ -1,214 +1,30 @@
...
+        Knight.knights_for_battle.get("Arthur").prep_for_battle(),
+        Knight.knights_for_battle.get("Red Knight").prep_for_battle())
 
-        if "protection" in lancelot["potion"]["effect"]:
-            lancelot["protection"] += lancelot["potion"]["effect"]["protection"]
+    battles_results.update(first_battle_result)
 
-        if "hp" in lancelot["potion"]["effect"]:
-            lancelot["hp"] += lancelot["potion"]["effect"]["hp"]
+    battles_results.update(second_battle_result)
```
LiudmylaKulinchenko, line 28: u can just use a key here for update a dict. `update` we use to merge dicts, you don't need it here

app/main.py
```diff
@@ -1,214 +1,30 @@
...
+        Knight.knights_for_battle.get("Lancelot").prep_for_battle(),
+        Knight.knights_for_battle.get("Mordred").prep_for_battle())
 
-    # apply potion if exist
-    if lancelot["potion"] is not None:
-        if "power" in lancelot["potion"]["effect"]:
-            lancelot["power"] += lancelot["potion"]["effect"]["power"]
+    second_battle_result = battle_arena.knight_battle(
+        Knight.knights_for_battle.get("Arthur").prep_for_battle(),
+        Knight.knights_for_battle.get("Red Knight").prep_for_battle())
```
LiudmylaKulinchenko, line 24: No need to use get here, you can access knight by key. It will increase readability, you anyway don't want knight to be None

LiudmylaKulinchenko (changes requested):

app/main.py
Naz-iv: fixed

app/main.py
Naz-iv: I decided to use update since my battle_results function returns dictionary and I didn't want to manually type keys/values for each knight, but can change it if necessary

app/main.py
Naz-iv: done

app/main.py
Naz-iv: since I don't use knight anywhere else apart from this code I didn't want to create additional variables. Code seamed to be simple enough to understand, but can change it if necessary

fsocie7y (approved):
well done ^_^

PR #437, opened 2023-10-06
Solution

prochigor:


Familenko (changes requested):
You are my opposite, too many catalogs. A separate module for one class with 4 lines is too much))

app/main.py
```diff
@@ -1,214 +1,31 @@
...
+    red_knight = Preparation.init_knight("red_knight", knights_config)
 
-    # check if someone fell in battle
-    if arthur["hp"] <= 0:
-        arthur["hp"] = 0
+    # equip knights
+    Action.preparation(lancelot)
+    Action.preparation(arthur)
+    Action.preparation(mordred)
+    Action.preparation(red_knight)
```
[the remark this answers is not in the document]
prochigor: OK, I have added two loops

actions/action.py
```diff
@@ -0,0 +1,22 @@
...
+        Preparation.apply_armour(knight)
+        Preparation.apply_weapon(knight)
+        Preparation.apply_potion(knight)
+
+    @staticmethod
+    def battle(first_knight: Knight, second_knight: Knight) -> None:
+        first_knight.hp -= second_knight.power - first_knight.protection
+        second_knight.hp -= first_knight.power - second_knight.protection
+
+        if first_knight.hp <= 0:
```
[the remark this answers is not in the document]
prochigor: Yes, I miss that, Done

knights/knight.py
```diff
@@ -0,0 +1,30 @@
...
+        self.armour = armour
+        self.weapon = weapon
+        self.potion = potion
+        self.protection = protection
+
+    def __str__(self) -> str:
+        return f"name: {self.name}, power: {self.power}, " \
+               f"hp: {self.hp}, protect: {self.protection}, " \
+               f"armour: {self.armour}, " \
+               f"weapon: {self.weapon}, potion: {self.potion}"
```
fsocie7y, line 30: Try to avoid using \, use ( ) instead

fsocie7y (changes requested):

app/main.py
```diff
@@ -1,214 +1,31 @@
...
+from app.data import knights_info
 
-    # check if someone fell in battle
-    if lancelot["hp"] <= 0:
-        lancelot["hp"] = 0
 
-    if mordred["hp"] <= 0:
-        mordred["hp"] = 0
+def battle(knights_config: dict) -> dict:
+    # init knights
```
LiudmylaKulinchenko, line 7: You can skip adding comments here, the logic is clear

app/main.py
```diff
@@ -1,214 +1,31 @@
...
+    knights = []
+    for knight in knights_config:
+        knight = Preparation.init_knight(knight, knights_config)
+        knights.append(knight)
```
LiudmylaKulinchenko, line 11: Use list comprehension

items/items_init.py
```diff
@@ -0,0 +1,28 @@
...
+class InitItems:
+
+    @staticmethod
+    def init_weapon(name: str, config: dict) -> Weapon:
+        weapon_name = config.get(name).get("weapon").get("name")
+        power = config.get(name).get("weapon").get("power")
+        return Weapon(weapon_name, power)
+
+    @staticmethod
+    def init_armour(name: str, config: dict) -> list[Armour]:
```
LiudmylaKulinchenko, line 15: Its not really initialization, but creation, because you create Armour object here

items/items_init.py
```diff
@@ -0,0 +1,28 @@
...
+        equipment = []
+        for part_arm in config.get(name).get("armour"):
+            part = part_arm.get("part")
+            protect = part_arm.get("protection")
+            equipment.append(Armour(part, protect))
+        return equipment
```
LiudmylaKulinchenko, line 21: You can use unpacking `**` and list comprehension here. You can even skip using `get` to access config values. You anyway don't want them to be None, so you can skip it here

items/items_init.py
```diff
@@ -0,0 +1,28 @@
...
+
+    @staticmethod
+    def init_potion(name: str, config: dict) -> Potion:
+        if config.get(str(name)).get("potion"):
+            potion_name = config.get(str(name)).get("potion").get("name")
+            effect = config.get(str(name)).get("potion").get("effect")
+            return Potion(potion_name, effect)
```
LiudmylaKulinchenko, line 28: Same here about get method. You can skip it
`config.get(str(name)).get("potion").get("name")` for exaple, if there is no str(name) if config, you will get None
And then your action will look like this `None.get("potion").get("name")`
So you will get the error `AttributeError: 'NoneType' object has no attribute 'get'`

LiudmylaKulinchenko: Fix it everywhere. It will increase readability and you will get just KeyError which will be easier to understand

knights/knight_preparation.py
```diff
@@ -0,0 +1,33 @@
...
+    @staticmethod
+    def apply_weapon(knight: Knight) -> None:
+        knight.power += knight.weapon.power
+
+    @staticmethod
+    def apply_potion(knight: Knight) -> None:
+        if knight.potion:
+            knight.protection += knight.potion.effect.get("protection", 0)
+            knight.hp += knight.potion.effect.get("hp", 0)
+            knight.power += knight.potion.effect.get("power", 0)
```
LiudmylaKulinchenko, line 21: Here is a good example of using `get`

items/items_init.py
```diff
@@ -0,0 +1,28 @@
+from items.weapon import Weapon
+from items.armour import Armour
+from items.potion import Potion
+
+
+class InitItems:
```
LiudmylaKulinchenko, line 6: Not really descriptive name for a class. You can add all this methods to the Weapon/Armour/Potion classes, it will look much more better and understandable. Don't forget about OOP

app/main.py
prochigor: Yes, it looks better)

knights/knight_preparation.py
```diff
@@ -0,0 +1,35 @@
...
+    def init_knight(name: str, knights_config: dict) -> Knight:
+        knight = Knight(
+            name.replace("_", " ").title(),
+            knights_config[name.lower()]["power"],
+            knights_config[name.lower()]["hp"],
+            Armour.create_armour(name, knights_config[name]),
+            Weapon.create_weapon(name, knights_config[name]),
+            Potion.crate_potion(name, knights_config[name])
+        )
+        return knight
```
LiudmylaKulinchenko, line 35: This method is more about Knight, as for me, not preparation. Also then u can make it classmethod, becuse u use class name here, so if it will be changed, you won't need to change it here
And again, it is not initialization, but creation. You should rename. It looks now like you rewrite your `Knight.__init__`

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Great job! 🎈

PR #438, opened 2023-10-07
Solution by The best in the world

SelFisher1488:


app/main.py
```diff
@@ -1,214 +1,24 @@
...
-    return {
-        lancelot["name"]: lancelot["hp"],
-        arthur["name"]: arthur["hp"],
-        mordred["name"]: mordred["hp"],
-        red_knight["name"]: red_knight["hp"],
-    }
-
-
-print(battle(KNIGHTS))
+    return {lancelot.name: lancelot.hp,
```
Andrii-Silenko, line 21: you can rewrite this with dict comprehension

app/knight_stats.py
```diff
@@ -0,0 +1,33 @@
+class Knight:
+    def __init__(self, equipment: dict) -> None:
```
Andrii-Silenko, line 2: maybe "name", "hp", "power" is not "equipment"

Andrii-Silenko (approved):

kostomeister (approved):
wow, amazing, how did you come up with this solution? Huge W, the best in the world at work

README.md
```diff
@@ -12,7 +12,7 @@ At first, you have knights stats as dictionary:
 ```python
 KNIGHTS = {
     "red_knight": {
-        "name": "Red Knight",
+        "name": "Red knights",
```
LiudmylaKulinchenko, line 15: You were not suppoused to change readme

tests/test_main.py
```diff
@@ -69,7 +69,7 @@ def base_knights_config():
             },
         },
         "red_knight": {
-            "name": "Red Knight",
+            "name": "Red knights",
```
LiudmylaKulinchenko, line 72: Don't change tests

app/knight_stats.py
LiudmylaKulinchenko: I agree 👍 Implicit expectiations, when you create Knight object, you don't know what exactly should be in equipment. And is `name` and equipmet?
To understand that I will need to go in this module and read the class logic. Change that

app/knight_stats.py
```diff
@@ -0,0 +1,33 @@
...
+        self.name = equipment.get("name")
+        self.hp = equipment.get("hp")
+        self.power = equipment.get("power")
+        self.protection = 0
+        self.armour = equipment.get("armour")
+        self.weapon = equipment.get("weapon")
+        self.potion = equipment.get("potion")
+
+    def apply_potion(self) -> None:
+        for stat, value in self.potion.get("effect").items():
```
LiudmylaKulinchenko, line 12: Let's imagine that potion doesn't have an "effect" key. Then `self.potion.get("effect")` will give you None. And it will lead to `None.items()` --> `AttributeError: 'NoneType' object has no attribute 'items'`
So It will be much more readable and easier to debug if you will access by key, without using `get`

LiudmylaKulinchenko: Fix everywhere

app/main.py
```diff
@@ -1,214 +1,24 @@
...
 
 
-def battle(knightsConfig):
-    # BATTLE PREPARATIONS:
+def battle(knights_config: dict) -> dict:
+    knights = knights_config
 
-    # lancelot
-    lancelot = knightsConfig["lancelot"]
+    knights_in_class = [Knight(knights.get(knight)) for knight in knights]
```
LiudmylaKulinchenko, line 8: Change name to more self-descriptive. You can name it `knights`

app/main.py
```diff
@@ -1,214 +1,24 @@
...
-    }
-}
+from app.knight_stats import Knight
+from app.battle_func import battle_action
 
 
-def battle(knightsConfig):
-    # BATTLE PREPARATIONS:
+def battle(knights_config: dict) -> dict:
+    knights = knights_config
```
LiudmylaKulinchenko, line 6: Do not rename your parameter here

LiudmylaKulinchenko (changes requested):

app/battle_func.py
LiudmylaKulinchenko, line 1: Rename this file

LiudmylaKulinchenko (changes requested):

app/main.py
LiudmylaKulinchenko: I agree, change that

LiudmylaKulinchenko (changes requested):

PR #439, opened 2023-10-08
Solution

Andrii-Silenko:


kostomeister (approved):
everything looks fine, but add return annotation to function in main.py

kostomeister (approved):
great

tests/test_main.py
```diff
@@ -1,6 +1,6 @@
 import pytest
 
-from app.main import battle
+from app.main import tournament
```
LiudmylaKulinchenko, line 3: Don't change tests

app/battle/battle.py
```diff
@@ -0,0 +1,12 @@
+from app.participants.knight import Knight
+
+
+def battle(*args: list[Knight, Knight]) -> None:
```
LiudmylaKulinchenko, line 4: You don't need to use args here

app/participants/knight.py
```diff
@@ -0,0 +1,23 @@
+class Knight:
+    def __init__(self, knight_config: dict) -> None:
```
LiudmylaKulinchenko, line 2: Implicit expectiations, when you create Knight object, you don't know what exactly should be in `knight_config`
To understand that I will need to go in this module and read the class logic. Change that

LiudmylaKulinchenko (changes requested):

PR #440, opened 2023-10-08
py-knights-fighting Solution

RVChornyy:


YuliaHladyshkevych (approved):

app/knights.py
```diff
@@ -0,0 +1,38 @@
+class Knights:
+    # This constructor is creating a knight (Knights class instance),
+    # which is ready for the battle
+    # (armoured and potion applied (if available)).
```
LiudmylaKulinchenko, line 4: Simplify the description and make it a docstring instead of a comment

LiudmylaKulinchenko: By simplify i mean don't add that knight is Knights class instance, it is obvious. So you don't need it in class description

app/knights.py
```diff
@@ -0,0 +1,38 @@
...
+        if armour:
+            for item in armour:
+                self.protection += item.get("protection")
+        extra_power = 0
+        if potion:
+            self.hp = hp + potion.get("effect").get("hp")
+            extra_power = potion.get("effect").get("power")
+            if potion.get("effect").get("protection"):
+                self.protection += potion.get("effect").get("protection")
+        self.power = power + extra_power + weapon.get("power")
```
LiudmylaKulinchenko, line 25: Add methods for this. Or even create another modules and classes for each object (armour, potion...)

app/knights.py
```diff
@@ -0,0 +1,38 @@
+class Knights:
```
LiudmylaKulinchenko, line 1: Rename this class. You create one Knight here, not Knights, so the name is wrong

app/main.py
```diff
@@ -86,129 +89,20 @@
...
+def battle(competitors: dict) -> dict:
+    # create knights as class "Knights" instances
+    fighters_list = [Knights(knight.get("name"),
+                             knight.get("hp"),
+                             knight.get("power"),
+                             knight.get("armour"),
+                             knight.get("weapon"),
+                             knight.get("potion"))
+                     for knight in competitors.values()]
+    # apply class "Knights" method "fight" to simulate the battle
```
LiudmylaKulinchenko, line 101: Remove comments, logic is understandable without them

app/main.py
```diff
@@ -86,129 +89,20 @@
...
-        red_knight["name"]: red_knight["hp"],
-    }
+def battle(competitors: dict) -> dict:
+    # create knights as class "Knights" instances
+    fighters_list = [Knights(knight.get("name"),
+                             knight.get("hp"),
+                             knight.get("power"),
+                             knight.get("armour"),
+                             knight.get("weapon"),
+                             knight.get("potion"))
```
LiudmylaKulinchenko, line 99: I can see in your Knights class that you don't want any of those attributes to be None, besides potion
Then why do you use `get` here? 
Also u can simplify it by using unpacking `**`

LiudmylaKulinchenko (changes requested):

PR #443, opened 2023-10-09
Solution

YuliaHladyshkevych:


Daniil-Pankieiev (approved):

app/battle/knight_creator.py
```diff
@@ -0,0 +1,85 @@
...
+        """
+        Args:
+            knight_nik (str):
+            The nickname of the Knight.
+            knight_params (dict):
+            The dictionary containing the parameters of the Knight.
+            list_order (int):
+            The position of the Knight in the list.
+        Returns:
+            object: The Knight object.
```
LiudmylaKulinchenko, line 19: You don't need to add explanation when login is understandable just by reading the code. Remove it

LiudmylaKulinchenko: Fix everywhere

app/battle/knight_creator.py
```diff
@@ -0,0 +1,85 @@
+class Knight:
+
+    def __init__(
+            self,
+            knight_nik: str,
```
LiudmylaKulinchenko, line 5: Don't need to add `knight_` part everywhere. It is easy to understand, because it is Knight object attribute
Also, from the description you know that here u use `name`, not a `nickname`. So change that

app/battle/knight_creator.py
```diff
@@ -0,0 +1,85 @@
...
+        self.name = knight_params["name"]
+        self.power = knight_params["power"]
+        self.hp = knight_params["hp"]
+        self.armour = self.full_armour(knight_params["armour"])
+        self.weapon = knight_params["weapon"]
+        self.potion = knight_params["potion"]
+        self.list_order = list_order
+
+    def __repr__(self) -> str:
+        # Returns a string representation of the knight.
```
LiudmylaKulinchenko, line 32: You don't need these comments

app/battle/knight_creator.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def full_armour(self, armour: list) -> int:
+        """
+        Calculates the total armour value
+        by adding up the protection values of each armour part.
+        """
+        output_armour = 0
+        for part in armour:
+            output_armour += part["protection"]
+        return output_armour
```
LiudmylaKulinchenko, line 54: Use sum instead

app/battle/knight_creator.py
```diff
@@ -0,0 +1,85 @@
...
+                if key == "protection":
+                    self.armour += value
+
+    def attack(self) -> int:
+        """
+        Calculates the Knight's attack value as the sum
+        of their power and weapon power.
+        """
+
+        return sum([self.power, self.weapon["power"]])
```
LiudmylaKulinchenko, line 74: sum works with generators so [] are not needed

app/main.py
```diff
@@ -1,214 +1,32 @@
...
-        lancelot["name"]: lancelot["hp"],
-        arthur["name"]: arthur["hp"],
-        mordred["name"]: mordred["hp"],
-        red_knight["name"]: red_knight["hp"],
-    }
-
-
-print(battle(KNIGHTS))
+# Print the battle simulation using the KNIGHTS dictionary
+print(dict(battle(KNIGHTS)))
```
LiudmylaKulinchenko, line 32: Remove print

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done 💎

PR #444, opened 2023-10-09
Solution

petrykivd:


app/battle/battle_starter.py
```diff
@@ -0,0 +1,30 @@
+from app.knights.knight_creation import Knight
+
+
+class Battle:
+    def __init__(self, knights_config: dict) -> None:
+        self.knights = []
+        for name in knights_config.keys():
+            self.knights.append(Knight(knights_config[name]))
```
LiudmylaKulinchenko, line 8: You can use list comprehension here. If you think it looks messy, you can create another method for that

tests/test_main.py
```diff
@@ -178,5 +178,4 @@ def test_arthur_and_lancelot_have_potion(base_knights_config):
         "Arthur": 60,
         "Mordred": 10,
         "Red Knight": 0,
-    }
-
+    }
```
LiudmylaKulinchenko, line 181: Add a blank line

app/main.py
```diff
@@ -86,129 +87,5 @@
...
-        lancelot["name"]: lancelot["hp"],
-        arthur["name"]: arthur["hp"],
-        mordred["name"]: mordred["hp"],
-        red_knight["name"]: red_knight["hp"],
-    }
-
-
-print(battle(KNIGHTS))
+def battle(knightsconfig: dict) -> dict:
+    return Battle(knightsconfig).battle()
```
LiudmylaKulinchenko, line 91: You should not create Battle class, because according to the task, you need `battle` as a function

LiudmylaKulinchenko (changes requested):

tests/test_main.py
```diff
@@ -179,4 +179,4 @@ def test_arthur_and_lancelot_have_potion(base_knights_config):
         "Mordred": 10,
         "Red Knight": 0,
     }
-
+
```
LiudmylaKulinchenko, line 182: This line is not blank

LiudmylaKulinchenko (approved):
Good job! 🐣

PR #448, opened 2023-10-10
Solution

sakhaline:


petrykivd (approved):

app/knight.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def apply_weapon(self) -> None:
+        self.power += self.weapon.power
+
+    def apply_potion(self) -> None:
+        if self.potion:
+            for stat, value in self.potion.effect.items():
+                self.__dict__[stat] += value
+
+    def fight(self, opponent: "Knight") -> None:
```
fsocie7y, line 43: u should use type annotation without quotes)

fsocie7y (changes requested):

app/knight.py
sakhaline: How can i do this? If I use opponent as Knight type in Knight class?

LiudmylaKulinchenko (approved):
Good job 😸

PR #450, opened 2023-10-12
Refactored code

DanSheremeta:
Implemented class-base solution of task.

### DIFF
--- a/app/knight_fighting/knight.py
+++ b/app/knight_fighting/knight.py
@@ -1,7 +1,13 @@
 class Knight:
-    def __init__(self, name: str, power: int,
-                 hp: int, armour: list[dict | None],
-                 weapon: dict, potion: dict = None) -> None:
+    def __init__(
+            self,
+            name: str,
+            power: int,
+            hp: int,
+            armour: list[dict | None],
+            weapon: dict,
+            potion: dict = None
+    ) -> None:
         self.name = name
         self.power = power
         self.hp = hp


app/knight_fighting/battle.py
```diff
@@ -0,0 +1,26 @@
...
+        self.first.hp -= self.second.power - self.first.protection
+        self.second.hp -= self.first.power - self.second.protection
+
+        if self.first.hp <= 0:
+            self.winner = self.second.name
+            self.first.hp = 0
+
+        if self.second.hp <= 0:
+            self.winner = self.first.name
+            self.second.hp = 0
```
YukihiroSM, line 20: There will be a case, when both of them will have <= 0 HP, but in this case the second one would be a winner. Don't think, that's correct.

app/knight_fighting/knight.py
```diff
@@ -0,0 +1,31 @@
+class Knight:
+    def __init__(self, name: str, power: int, hp: int,
+                 armour: list[dict | None], weapon: dict,
+                 potion: dict = None) -> None:
```
YukihiroSM, line 4: If you are separating lines in func/method definition - better to have one parameter in one line.

YukihiroSM (changes requested):

app/knight_fighting/knight.py
```diff
@@ -0,0 +1,31 @@
+class Knight:
+    def __init__(self, name: str, power: int,
+                 hp: int, armour: list[dict | None],
+                 weapon: dict, potion: dict = None) -> None:
```
NikOneZ1, line 4: If you are separating lines in func/method definition - better to have one parameter in one line like this:
```suggestion
    def __init__(
        self, 
        name: str, 
        power: int,
        hp: int, 
        armour: list[dict | None],
        weapon: dict, 
        potion: dict = None
    ) -> None:
```

NikOneZ1 (changes requested):

EdAlekseiev (approved):
Good job

PR #456, opened 2023-10-19
Solution

CrazyHilly:


app/modules/knight.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class Knight:
+    def __init__(
+            self,
+            name: str = "",
+            power: int = 0,
+            hp: int = 0,
+            armour: list[dict] = [],
+            weapon: dict = {},
```
EdAlekseiev, line 11: dict and list cannot be default arguments, because they are mutable.
https://www.codecademy.com/learn/learn-intermediate-python-3/modules/int-python-function-arguments/cheatsheet

EdAlekseiev (changes requested):

app/modules/knight.py
YukihiroSM: Not fixed.

YukihiroSM (changes requested):

app/modules/knight.py
CrazyHilly: Forgot to git push, sorry about that.

EdAlekseiev (approved):
Good job

PR #458, opened 2023-10-22
Creative solution with self logic

Serhii-Hrynyshyn:


app/knight.py
```diff
@@ -0,0 +1,136 @@
...
+        self.hp = hp
+
+        Knight.knights.add(self)
+
+    knights = set()
+
+    def take_a_weapon(self, name: str) -> object:
+        # add power from take a weapon
+        self.power += weapon.value[name][0]["power"]
+        return weapon.value[name][0]["power"]
```
ArsenPidhoretskyi, line 19: Why you return this?

app/knight.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def dress_on_armors(self, name: str) -> list:
+        # add hp from dress on an armor
+        total_armor = []
+        for el in armor.value[name]:
+            self.hp += el["protection"]
```
ArsenPidhoretskyi, line 25: store protection into protection attribute

app/knight.py
```diff
@@ -0,0 +1,136 @@
...
+                    potion.value[name]["effect"]["power"])
+            if "protection" in potion.value[name]["effect"]:
+                self.hp += potion.value[name]["effect"]["protection"]
+                _pn = "Protection: " + str(potion.value[name]["effect"][
+                                               "protection"])
+            if "hp" in potion.value[name]["effect"]:
+                self.hp += potion.value[name]["effect"]["hp"]
+                _hp = "HP: " + str(
+                    potion.value[name]["effect"][
+                        "hp"])
```
ArsenPidhoretskyi, line 47: Use loops instead of duplicate code
Also use getatte, setattr

app/knight.py
```diff
@@ -0,0 +1,136 @@
...
+              "{0:^67}".format(f"Power: {second.power}"))
+
+    @classmethod
+    def versus(cls, first_fighter: str, second_fighter: str) -> None:
+        # Choices first and second fighters between all knights
+        for el in Knight.knights:
+            if el.name == first_fighter:
+                first_fighter = el
+            if el.name == second_fighter:
+                second_fighter = el
```
ArsenPidhoretskyi, line 65: In this case better to use knights as dict

ArsenPidhoretskyi (changes requested):
Why you remove battle function from main

PR #459, opened 2023-10-24
Solution

OlehShumov:


app/apply.py
```diff
@@ -0,0 +1,24 @@
+def apply_everything(name_of_knight: str, knights_config: dict) -> dict:
```
LiudmylaKulinchenko, line 1: `everything` is not really descriptive. Change it. You can create different functions, classes to make a logic much more clear

LiudmylaKulinchenko: Also change a module name

app/main.py
```diff
@@ -202,7 +116,7 @@ def battle(knightsConfig):
     if red_knight["hp"] <= 0:
         red_knight["hp"] = 0
 
-    # Return battle results:
+    # Return battle results as a dictionary:
```
LiudmylaKulinchenko, line 119: Remove comments. Fix everywhere

LiudmylaKulinchenko (changes requested):

PR #463, opened 2023-10-29
Solution

spryima:


app/battle/knights_creation.py
```diff
@@ -0,0 +1,30 @@
+from app.classes.weapon import Weapon
+from app.classes.armour import Armour
+from app.classes.potion import Potion
+from app.classes.knight import Knight
+
+
+def creating_knights(knights_config: dict) -> dict[str: Knight]:
+    knights = {}
+
+    for key, knight in knights_config.items():
```
ArsenPidhoretskyi, line 10: rename key to name

app/battle/knights_creation.py
```diff
@@ -0,0 +1,30 @@
...
+
+        if knight["potion"]:
+            knights[key].potion = Potion(
+                knight["potion"]["name"],
+                knight["potion"]["effect"].get("power", 0),
+                knight["potion"]["effect"].get("hp", 0),
+                knight["potion"]["effect"].get("protection", 0)
+            )
+        else:
+            knights[key].potion = Potion("None")
```
ArsenPidhoretskyi, line 28:
```suggestion
            knights[key].potion = None
```
Bad practice to create object Potion("None")

app/classes/knight.py
```diff
@@ -0,0 +1,38 @@
...
+        self.total_hp = 0
+        self.total_power = 0
+        self.total_protection = 0
+
+    def ready_for_battle(self) -> None:
+        self.total_hp = self.hp + self.potion.hp
+        self.total_power = self.power + self.weapon.power + self.potion.power
+        self.total_protection = self.potion.protection
+        for armour in self.armour:
+            self.total_protection += armour.protection
```
ArsenPidhoretskyi, line 34: In this case better to create @property for each variable instead of usage attribute

ArsenPidhoretskyi (changes requested):

PR #476, opened 2023-11-06
'solution'

xborismenx:


YukihiroSM (changes requested):
this solution looks ok, but there is some logic missmatch. It's strange, that knights are instances of Battle. What about creating Knight class and moving knights attributes there? Because Battle shouldn't create knights.

app/check_stats.py
```diff
@@ -0,0 +1,41 @@
...
+            potion_effect = self.knights["potion"]["effect"]
+            # if potion_hp, + on base hp
+            if "hp" in potion_effect:
+                self.hp += potion_effect["hp"]
+            # if potion_power, + on base power
+            if "power" in potion_effect:
+                self.power += potion_effect["power"]
+            # if potion_protection, + on base hp
+            if "protection" in potion_effect:
+                self.hp += potion_effect["protection"]
```
EdAlekseiev, line 28: I think, that you should separate this method to 3 methods. Method for calculating weapon, calculating armour and calculating potion effects.

EdAlekseiev (changes requested):

YukihiroSM (approved):
Nicely done!

PR #497, opened 2023-11-25
solve py-knights-fighting

zdmytryszyn:


app/battle/battle_start.py
```diff
@@ -0,0 +1,22 @@
...
+class Battle:
+    def __init__(self, knight_1: Knight, knight_2: Knight) -> None:
+        self.knight_1 = knight_1
+        self.knight_2 = knight_2
+
+    def battle_start(self) -> None:
+        self.knight_1.health_points -= \
+            self.knight_2.power - self.knight_1.protection
+        self.knight_2.health_points -= \
+            self.knight_1.power - self.knight_2.protection
```
vsmutok, line 15: better to use `()` then `\`

app/battle/battle_start.py
```diff
@@ -0,0 +1,22 @@
+from __future__ import annotations
```
vsmutok, line 1: remove redundant import

app/main.py
```diff
@@ -1,214 +1,56 @@
...
 
-        if "protection" in red_knight["potion"]["effect"]:
-            red_knight["protection"] += red_knight["potion"]["effect"]["protection"]
+        if knight.potion is not None:
+            potion = Potion(knight.potion["effect"])
+            potion.apply_potion(knight)
 
-        if "hp" in red_knight["potion"]["effect"]:
-            red_knight["hp"] += red_knight["potion"]["effect"]["hp"]
+    print(list_of_knights["lancelot"].name)
```
vsmutok, line 36: put it in the Knight class, so that it is called when a new knight is initialized

vsmutok (changes requested):

app/main.py
zdmytryszyn: when I put this part into the Knight class, the circular import occurs between modules knight/potion, weapon, armour. how to execute this part of code inside the knight module without circulat import error arising?

app/main.py
```diff
@@ -1,214 +1,39 @@
...
 
-    # apply armour
-    mordred["protection"] = 0
-    for a in mordred["armour"]:
-        mordred["protection"] += a["protection"]
+    list_of_knights = {}
 
-    # apply weapon
-    mordred["power"] += mordred["weapon"]["power"]
+    # KNIGHTS INITIATION
```
Dimosphen1, line 10: Remove this comment, there is no need for it.

app/contestants/list_of_contestants.py
```diff
@@ -0,0 +1,88 @@
+class Contestants:
```
Dimosphen1, line 1: There is no need to create a separate class to only store a variable. Make `KNIGHTS` a separate variable and import it directly where needed.

app/main.py
```diff
@@ -1,214 +1,39 @@
...
-    if arthur["hp"] <= 0:
-        arthur["hp"] = 0
-
-    if red_knight["hp"] <= 0:
-        red_knight["hp"] = 0
+    battle_2 = Battle(list_of_knights["arthur"], list_of_knights["red_knight"])
+    battle_2.battle_start()
+    battle_2.check_if_fell()
 
     # Return battle results:
```
Dimosphen1, line 32: Remove this comment and others in other files, there is no need for them.

app/equipment/weapon.py
```diff
@@ -0,0 +1,10 @@
+from app.contestants.knight import Knight
+
+
+class Weapon:
+    def __init__(self, weapon_power: int) -> None:
+        self.weapon_power = weapon_power
+
+    # apply weapon
+    def apply_weapon(self, knight: Knight) -> None:
+        knight.power += self.weapon_power
```
Dimosphen1, line 10: An overall comment regarding the `equipment` module, you create separate classes for each piece of equipment that contain a single method applied to a knight instance. It is not the right approach because when you have a group of methods that act exclusively on the same instance it is the first sign to consider placing them in a class to perform actions over the instance directly. With this approach, you'll avoid a circular import of the `Knight` class and will have a general understanding of all actions that could be applied over a knight instance in terms of equipment.

app/equipment/potion.py
```diff
@@ -0,0 +1,18 @@
...
+    def apply_potion(self, knight: Knight) -> None:
+        if knight.potion is not None:
+            if "power" in self.effect:
+                knight.power += self.effect["power"]
+
+            if "protection" in self.effect:
+                knight.protection += self.effect["protection"]
+
+            if "hp" in self.effect:
+                knight.health_points += self.effect["hp"]
```
Dimosphen1, line 18: Try to avoid using multiple `if` conditions and use for loop instead according to the [checklist](https://github.com/zdmytryszyn/py-knights-fighting/blob/develop/checklist.md#dont-repeat-yourself).

Dimosphen1 (changes requested):

app/equipment/weapon.py
zdmytryszyn: Placed all separate classes, connected with equipment, into the same class. Still don't get how to avoid circular import. Equipment class still have to import from Knight class to perform actions over it. Vice versa, Knight class needs to import from Equipment class in order to apply equipment every time the Knight instance is created.

Polyakiv-Andrey (approved):
Good job!

PR #499, opened 2023-11-28
'Solution' py-knights-fighting

yura-push:


app/knight/knight_class.py
```diff
@@ -0,0 +1,47 @@
...
+
+                if "hp" in value["effect"]:
+                    self.hp += value["effect"]["hp"]
+
+    def extract_potion(self, knights: dict) -> dict:
+        return {
+            knights[knight]["name"]: knights[knight]["potion"]
+            for knight in knights
+            if knights[knight]["name"] == self.name
+        }
```
Dimosphen1, line 43: Rather than iterate through all knights, provide the necessary data regarding the knight's armor, weapon, and potion during the initialization, so that you could call these methods in the `__init__` and they already will do all the job without the need for multiple iteration.

app/battle_preparations/prepare_to_battle.py
```diff
@@ -0,0 +1,21 @@
...
+    ]
+
+    for knight_inst in knights_instances:
+        knight_inst.apply_armor(knights)
+
+    for knight_inst in knights_instances:
+        knight_inst.equip_weapon(knights)
+
+    for knight_inst in knights_instances:
+        knight_inst.use_potion(knights)
```
Dimosphen1, line 19: Avoid doing multiple loopings, provide and apply all data during the initialization.

app/knight/knight_class.py
```diff
@@ -0,0 +1,47 @@
...
+        for key, value in potion.items():
+            if value is not None:
+                if "power" in value["effect"]:
+                    self.power += value["effect"]["power"]
+
+                if "protection" in value["effect"]:
+                    self.protection += value["effect"]["protection"]
+
+                if "hp" in value["effect"]:
+                    self.hp += value["effect"]["hp"]
```
Dimosphen1, line 36: Think of an implementation that would eliminate the need for three similar if conditions. You could replace them with a loop instead.

app/main.py
```diff
@@ -1,214 +1,39 @@
...
+    else:
+        lancelot.hp = mordreds_attack
 
-        if "hp" in mordred["potion"]["effect"]:
-            mordred["hp"] += mordred["potion"]["effect"]["hp"]
+    arthur_attack = arthur.attack(red_knight.hp, red_knight.protection)
+    if arthur_attack <= 0:
+        red_knight.hp = 0
+    else:
+        red_knight.hp = arthur_attack
```
Dimosphen1, line 26: You could separate this logic and place it in the attack so that there would be no need for code duplication on lines 29 up to 32 and similar code above.

app/battle_preparations/prepare_to_battle.py
```diff
@@ -0,0 +1,21 @@
+from app.knight.knight_class import Knight
+
+
+def preparations(knights: dict) -> list:
```
Dimosphen1, line 4: Rename the function name, it should use verbs and represent action. You could use the name `prepare` instead.

Dimosphen1 (changes requested):

pavlejviki (changes requested):
Fix all the comments

ArsenPidhoretskyi (changes requested):

PR #529, opened 2024-01-10
Solution

aLEKS-e3:


app/knight_class.py
isikach, line 1: May be it can be better if you add results as Class Attribute

isikach (approved):

app/knight_class.py
aLEKS-e3: thank you, that's a great idea!

app/preparing_knights.py
```diff
@@ -0,0 +1,12 @@
+def battle_preparation(knights: dict) -> dict:
+    for knight in knights.values():
+        knight["protection"] = sum(armour["protection"]
+                                   for armour in knight["armour"])
+        knight["power"] += knight["weapon"]["power"]
+
+        if knight["potion"] is not None:
+            for stat in ["power", "protection", "hp"]:
+                if stat in knight["potion"]["effect"]:
+                    knight[stat] += knight["potion"]["effect"][stat]
```
NEkropo1, line 10: I guess better to create potion class, so we can now if there's some effects, if we can modify them, and to ensure there won't be key errors.

app/main.py
```diff
@@ -86,129 +90,19 @@
...
-        if "hp" in red_knight["potion"]["effect"]:
-            red_knight["hp"] += red_knight["potion"]["effect"]["hp"]
-
+    knights = {name: Knight(prepared_knights[name])
+               for name in
+               ["lancelot", "arthur", "mordred", "red_knight"]}
     # -------------------------------------------------------------------------------
     # BATTLE:
+    knights["lancelot"].fight(knights["mordred"])
+    knights["arthur"].fight(knights["red_knight"])
```
NEkropo1, line 103: It might be more flexible to not hardcode the knight names inside the battle function. Consider passing the matchups as an argument or designing a more dynamic way to handle fights.

NEkropo1 (changes requested):

app/arena.py
```diff
@@ -0,0 +1,30 @@
...
+
+        if knight1.hp > 0 and knight2.hp > 0:
+            knight1.hp = max(
+                knight1.hp - (knight2.power - knight1.protection), 0
+            )
+            knight2.hp = max(
+                knight2.hp - (knight1.power - knight2.protection), 0
+            )
+        else:
+            print("Dead knights can't fight!")
```
NEkropo1, line 21: Better to make some function like hit(), where first knight hits second, so it's more obvious, and leave that logic inside knight class, while here we only call knight_1.hit(knight(2) and knight_2.hit*knight(1) so we can only see the result.
Also if this message is looking for unexpected behavior, better to Raise an error here, so we know, we have problem somewhere else, since we shouldn't let knights with hp <= 0 to fight. Otherwise we just leave unnecessary print in app state.

app/knight.py
```diff
@@ -0,0 +1,36 @@
...
+
+        if potion_data := knight_data["potion"]:
+            potion = Potion(**potion_data["effect"])
+            potion.apply(knight_data)
+
+    @classmethod
+    def get_statistics(cls, knights: dict) -> dict:
+        for config in knights.values():
+            cls.__knights_statistics[config.name] = config.hp
+        return cls.__knights_statistics
```
NEkropo1, line 36: still can't get what's an idea of this? name `get_statistics` supposes to get info about the statistics, while this method rewrites knights hp. Maybe you should make getter and setter and use them in different scenarios, or make some `update` method?)

app/potion.py
```diff
@@ -0,0 +1,28 @@
...
+        for stat, boost in self._get_boosts().items():
+            knight_data[stat] = knight_data.get(stat, 0) + boost
+        return knight_data
+
+    def _get_boosts(self) -> dict:
+        return {
+            "hp": self.hp,
+            "power": self.power,
+            "protection": self.protection
+        }
```
NEkropo1, line 28: in case of private class it's more like `_get_attributes`, since we're talking about this class, not knight.

app/main.py
```diff
@@ -1,107 +1,26 @@
...
+        pairs: list | tuple = None
+) -> dict:
     # BATTLE PREPARATIONS:
-    prepared_knights = battle_preparation(knights_config)
+    knights = Arena.equip_knights(knights_config)
+
+    battle_pairs = pairs or [
+        ["lancelot", "mordred"],
+        ["arthur", "red_knight"],
+    ]
```
NEkropo1, line 16: Best practice is to make predefined pairs, where you load these data, like from some config file or env.

app/knight.py
```diff
@@ -0,0 +1,36 @@
+from __future__ import annotations
+from app.potion import Potion
+
+
+class Knight:
+    __knights_statistics = {}
+
+    def __new__(cls, knight: dict, *args, **kwargs) -> Knight:
+        cls._prepare_for_battle(knight)
+        cls.__knights_statistics[knight.get("name")] = knight.get("hp")
```
NEkropo1, line 10: this could be just static method, since you don't need to save this knights_statistics in class, since you overwright it, and you can simply work with Arena one, where you save overall statistics, and can just update certain knight info for every knight, while having links in arena on them.

NEkropo1 (changes requested):

app/knight.py
aLEKS-e3: how can it be a static method since before creating a knight I'm preparing him? or did I get it wrong?

app/main.py
aLEKS-e3: ok, I'll pay attention to this, but I won't rewrite it so as not to get too deep and bother you)

PR #532, opened 2024-01-13
Solution

isikach:


aLEKS-e3 (approved):
I like this solution!

app/Kings_fighting/kings_start_position.py
```diff
@@ -0,0 +1,19 @@
+from __future__ import annotations
+from app.Kings_fighting.addition_tools import magic_upgrades
+
+
+class King:
+    # It can be useful to check correct start parameters
```
NEkropo1, line 6: not sure if you need this comment. Like it should be either implemented or not, or, at least, in TODO

app/Kings_fighting/kings_start_position.py
```diff
@@ -0,0 +1,19 @@
+from __future__ import annotations
+from app.Kings_fighting.addition_tools import magic_upgrades
+
+
+class King:
```
NEkropo1, line 5: Why King if we have knights fighting?) I mean it's up to you, just seem not intuitive, as for me.

NEkropo1 (approved):
Good job!

PR #533, opened 2024-01-13
Solution

ArturPoltser:


MaxymChyncha (approved):
Good job!

app/classes/knight.py
```diff
@@ -1,22 +1,17 @@
...
-        self.hp = knight["hp"]
-        self.armour = knight["armour"]
-        self.weapon = knight["weapon"]
-        self.potion = knight["potion"]
+        self.name = knight.get("name")
+        self.power = knight.get("power")
+        self.hp = knight.get("hp")
+        self.armour = knight.get("armour")
+        self.weapon = knight.get("weapon")
+        self.potion = knight.get("potion")
```
NEkropo1, line 8: I'd prefer something more explicit, like knight_dict, since it's not an object atm, so it's like same object in one scope which can sometime trick you.

NEkropo1 (approved):
Good job!

PR #537, opened 2024-01-13
py-knights-fighting

Vladyslav-Pliashechnyk:


app/knight.py
```diff
@@ -0,0 +1,30 @@
...
+        self.name = knight["name"]
+        self.power = knight["power"]
+        self.hp = knight["hp"]
+        self.armour = knight["armour"]
+        self.weapon = knight["weapon"]
+        self.potion = knight["potion"]
+        self.protection = 0
+
+        # a knight prepares for battle
+        self.battle_preparations()
```
isikach, line 12: You can use this function to decorate init func

isikach (approved):

app/knight.py
```diff
@@ -0,0 +1,30 @@
+class Knight:
+    def __init__(self, knight: dict) -> None:
+        self.name = knight["name"]
+        self.power = knight["power"]
+        self.hp = knight["hp"]
+        self.armour = knight["armour"]
+        self.weapon = knight["weapon"]
+        self.potion = knight["potion"]
```
NEkropo1, line 8: that's pretty bad approach, since you should create knight not from dict, you can use some manager that translates dict to parameters, but not unpack dict like that.

app/knight.py
NEkropo1, line 12: not sure it should be in an init method, more logical that knight should prepare for a battle when he is going to a battle, not when he just initialized)

app/battle.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+def fight(first_knight: Knight, second_knight: Knight) -> None:
+    first_knight.hp -= second_knight.power - first_knight.protection
+    second_knight.hp -= first_knight.power - second_knight.protection
+
+    # check if someone fell in battle
+    for knight in (first_knight, second_knight):
+        if knight.hp <= 0:
+            knight.hp = 0
```
NEkropo1, line 11: this also could be some default check, or callback(advanced)

NEkropo1 (changes requested):

PR #547, opened 2024-01-20
solution

nicksetrakov:


app/knights.py
```diff
@@ -0,0 +1,49 @@
+from __future__ import annotations
+
+
+class Knights:
```
Sofbog, line 4: why is the class name plural?

Sofbog (changes requested):

app/main.py
```diff
@@ -1,214 +1,21 @@
...
+from app.battle import Battle
 
 
-def battle(knightsConfig):
-    # BATTLE PREPARATIONS:
+def battle(config: dict) -> dict:
+    lancelot = Knight.create_knight(config, "lancelot")
+    mordred = Knight.create_knight(config, "mordred")
+    arthur = Knight.create_knight(config, "arthur")
+    red_knight = Knight.create_knight(config, "red_knight")
```
YuriiKindrat, line 10: gj, but could u use loop for this ?

app/knights.py
```diff
@@ -0,0 +1,49 @@
...
+    ) -> None:
+        self.name = name
+        self.power = power
+        self.hp = hp
+        self.armour = armour
+        self.weapon = weapon
+        self.potion = potion
+        self.protection = 0
+
+    def is_die(self) -> bool:
```
YuriiKindrat, line 23: it is better to name it "is_dead"/"is_died"

YuriiKindrat (changes requested):

YuriiKindrat (approved):
LGTM!

PR #562, opened 2024-01-29
Solution

mvsydoruk:
Solution

### DIFF
--- a/app/knight/knight.py
+++ b/app/knight/knight.py
@@ -1,0 +1,43 @@
+from __future__ import annotations
+
+
+class Knight:
+    def __init__(
+            self,
+            name: str,
+            power: int,
+            hp: int,
+            armour: list,
+            weapon: dict,
+            potion: dict = None,
+            protection: int = 0
+    ) -> None:
+        self.name = name
+        self.power = power
+        self.hp = hp
+        self.armour = armour
+        self.weapon = weapon
+        self.potion = potion
+        self.protection = protection
+
+    def apply_armour(self) -> None:
+        self.protection = sum([a["protection"] for a in self.armour])
+
+    def apply_weapon(self) -> None:
+        self.power += self.weapon["power"]
+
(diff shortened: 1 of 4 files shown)


app/knight/knight.py
```diff
@@ -0,0 +1,43 @@
...
+        self.name = name
+        self.power = power
+        self.hp = hp
+        self.armour = armour
+        self.weapon = weapon
+        self.potion = potion
+        self.protection = protection
+
+    def apply_armour(self) -> None:
+        self.protection = sum([a["protection"] for a in self.armour])
```
pavlejviki, line 24: a is not a descriptive variable name

app/knight/knight.py
```diff
@@ -0,0 +1,43 @@
...
+        self.power += self.weapon["power"]
+
+    def apply_potion(self) -> None:
+        if self.potion is not None:
+            effect = self.potion["effect"]
+            if "hp" in effect:
+                self.hp += effect["hp"]
+            if "power" in effect:
+                self.power += effect["power"]
+            if "protection" in effect:
```
pavlejviki, line 36: instead of if conditions you can use get() method with a default value 0 if a certain key is not found in the dict

app/knight/knight.py
```diff
@@ -0,0 +1,43 @@
...
+        self.protection = sum([a["protection"] for a in self.armour])
+
+    def apply_weapon(self) -> None:
+        self.power += self.weapon["power"]
+
+    def apply_potion(self) -> None:
+        if self.potion is not None:
+            effect = self.potion["effect"]
+            if "hp" in effect:
+                self.hp += effect["hp"]
```
pavlejviki, line 33:
```suggestion
                self.hp += effect.get("hp", 0)
```

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #563, opened 2024-01-29
Solution

liudmylaperchevska:


Polyakiv-Andrey (changes requested):
1)Split this code into multiple files
2)Delete comments from code, it's better if you divide code on  functions

app/main.py
```diff
@@ -1,4 +1,6 @@
-KNIGHTS = {
+from app.battle.battle import battle
+
+knights = {
```
Dimosphen1, line 3: Place this variable in a separate file and name it `knight_config.py` to unload the `main.py` file. You could import this variable from the file as well. Besides that, it would be better to leave this variable in upper-case indicating that it is a constant.

app/battle/pre_battle.py
```diff
@@ -0,0 +1,17 @@
+def pre_battle(knights_one: dict) -> dict:
```
Dimosphen1, line 1: It would be better to use more descriptive naming for this function and file, e.g. `prepare_knights`.

Dimosphen1 (changes requested):

PR #565, opened 2024-01-30
Develop

MrsLFG:


app/main.py
```diff
@@ -86,123 +89,34 @@
...
-    if lancelot["hp"] <= 0:
-        lancelot["hp"] = 0
+    for knight in knights:
+        apply_armour(knight)
+        apply_weapon(knight)
+        apply_potion(knight)
 
-    if mordred["hp"] <= 0:
-        mordred["hp"] = 0
+    def apply_damage(attacker: dict, defender: dict) -> None:
```
vasylhnatiuk, line 105: don't write function in another function

vasylhnatiuk (changes requested):
Please create Class Knight and save all data about knights as attributes, and also use methods instead of functions

YukihiroSM (approved):
Looks good to me!

PR #570, opened 2024-02-02
Solution

vsupruniuk:


app/knight/knight.py
```diff
@@ -0,0 +1,62 @@
+from app.knight_equipment.armour import Armour
+from app.knight_equipment.potion import Potion
+from app.knight_equipment.weapon import Weapon
+
+
+class Knight:
+    def __init__(self, knight_config: dict) -> None:
```
LiudmylaKulinchenko, line 7: How about make init method get name, power, hp... one by one here in parameters, not whole knight_config dict?
It will make easier to understand required parameters to create a Knight without looking directly in this method
It'll also let you set default values easier

LiudmylaKulinchenko: After that you could use unpacking (using `**`) to create knights + list comprehension to create list of knights
`[Knight(**knight) for knight in knight_config]`

app/knight/knight.py
```diff
@@ -0,0 +1,62 @@
...
+
+        self.weapon = Weapon(
+            name=knight_config["weapon"]["name"],
+            power=knight_config["weapon"]["power"]
+        )
+
+        self.potion = Potion(
+            name=knight_config["potion"]["name"],
+            effect=knight_config["potion"]["effect"]
+        ) if knight_config.get("potion") else None
```
LiudmylaKulinchenko, line 27: You can also use unpacking for armour, weapon, potion

LiudmylaKulinchenko (changes requested):
Looks good! Made you some suggestions

LiudmylaKulinchenko (approved):
Looks good! 🔥

PR #572, opened 2024-02-04
solution

KovbasinskyV:


vasylhnatiuk (changes requested):
create class Knight and al needed methods in there to make your code more clear

PR #581, opened 2024-02-11
Add files via upload

opminen:


app/knights/preparation.py
```diff
@@ -0,0 +1,37 @@
...
+    def apply_preparations(self) -> None:
+        if self.armour is not None:
+            for item in self.armour:
+                self.protection += item["protection"]
+
+        self.power += self.weapon["power"]
+
+        if self.potion is not None:
+            potion_effect = self.potion["effect"]
+            self.power += potion_effect.get("power", 0)
```
vasylhnatiuk, line 35: use settatr and loop instead of duplicating

vasylhnatiuk (changes requested):

PR #582, opened 2024-02-11
Solution

IhorDmytriv:


app/knights/knight.py
```diff
@@ -0,0 +1,31 @@
...
+        self.protection = 0
+
+        for armour in self.armour:
+            self.protection += armour.get("protection")
+
+        self.power += self.weapon.get("power")
+
+        if self.potion:
+            for effect, value in self.potion.get("effect").items():
+                if effect == "hp":
```
vasylhnatiuk, line 21: use setattr and loop instead of many if

vasylhnatiuk (changes requested):

EdAlekseiev (approved):
Good job

PR #586, opened 2024-02-16
'Solution'

olessh:


vasylhnatiuk (changes requested):
don't create file for each function! instead of it create class Knight and save all needed methods there

app/main.py
```diff
@@ -86,121 +93,39 @@
...
-    red_knight["hp"] -= arthur["power"] - red_knight["protection"]
+    results_hp(arthur, red_knight)
 
     # check if someone fell in battle
-    if arthur["hp"] <= 0:
-        arthur["hp"] = 0
-
-    if red_knight["hp"] <= 0:
-        red_knight["hp"] = 0
+    check_fell(arthur)
```
vasylhnatiuk, line 127: try to avoid duplicating

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -86,129 +89,75 @@
...
+        self.hp -= other_knight.power - self.protection
+        other_knight.hp -= self.power - other_knight.protection
 
-    # apply weapon
-    mordred["power"] += mordred["weapon"]["power"]
+    def battle_pre(self, other_knight: Knight) -> None:
+        # BATTLE PREPARATIONS:
+        self.apply_armour()
+        self.apply_weapon()
+        self.apply_potion_if_exist()
```
vsmutok, line 131:
```suggestion
        for instance in (self, other_knight):
            # BATTLE PREPARATIONS:
            instance.apply_armour()
            instance.apply_weapon()
            instance.apply_potion_if_exist()
```

you can use a loop to avoid code duplication

app/main.py
```diff
@@ -1,3 +1,6 @@
+from __future__ import annotations
+
+
 KNIGHTS = {
     "lancelot": {
         "name": "Lancelot",
```
vsmutok, line 6: move dict with data to separate module

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #595, opened 2024-02-22
Solution

TIgor22:


app/class_knight.py
```diff
@@ -0,0 +1,26 @@
...
+        for element_armor in self.armour:
+            self.protection += element_armor["protection"]
+        return self.protection
+
+    def knight_takes_weapon(self) -> None:
+        self.power += self.weapon["power"]
+
+    def knight_drinks_the_potion(self) -> None:
+        if self.potion is not None:
+            for key, value in self.potion["effect"].items():
```
vasylhnatiuk, line 24: key, value are bad names. use more accurate

vasylhnatiuk (changes requested):

app/class_knight.py
TIgor22: ok i will try now

PR #596, opened 2024-02-23
Solution

mish-sk:


app/main.py
```diff
@@ -86,129 +91,36 @@
...
+    for knight in dictionary_of_knights:
+        dictionary_of_knights.get(knight).prepare_for_the_fight()
+        dictionary_of_knights.get(knight).apply_potion()
+
+    lancelot = dictionary_of_knights.get("lancelot")
+    mordred = dictionary_of_knights.get("mordred")
+    arthur = dictionary_of_knights.get("arthur")
+    red_knight = dictionary_of_knights.get("red_knight")
+
+    lancelot.fight_and_apply_damage(mordred)
```
vasylhnatiuk, line 116: use loop instead of duplicating

app/main.py
```diff
@@ -86,129 +91,36 @@
...
-    if red_knight["hp"] <= 0:
-        red_knight["hp"] = 0
-
-    # Return battle results:
     return {
-        lancelot["name"]: lancelot["hp"],
-        arthur["name"]: arthur["hp"],
-        mordred["name"]: mordred["hp"],
-        red_knight["name"]: red_knight["hp"],
+        lancelot.name: lancelot.hp,
```
vasylhnatiuk, line 122: use loop instead of duplicating

app/knights/knight.py
```diff
@@ -0,0 +1,40 @@
...
+        self.protection = 0
+
+    def prepare_for_the_fight(self) -> None:
+        for equipment in self.armour:
+            self.protection += equipment.protection
+
+        self.power += self.weapon.power
+
+    def apply_potion(self) -> None:
+        if hasattr(self.potion, "effect") and \
```
vasylhnatiuk, line 31: use () instead of \

vasylhnatiuk (changes requested):

app/main.py
mish-sk: As in this particular task, each knight has its own rival that they fighting and not fighting every other knight. I believe the loop would not be much more efficient then hardcoding the battle between them.

YukihiroSM (approved):
Looks good to me!

PR #601, opened 2024-02-27
Solution

Viacheslav2024:


app/knight/knight.py
```diff
@@ -0,0 +1,61 @@
...
+    def __apply_armour(self) -> None:
+        self.__protection = sum(item["protection"] for item in self.__armour)
+
+    def __apply_weapon(self) -> None:
+        self.__power += self.__weapon["power"]
+
+    def __apply_potion(self) -> None:
+        potion_effect = self.__potion["effect"]
+
+        if "power" in potion_effect:
```
vasylhnatiuk, line 37: instead of duplicating use loop and settattr

app/main.py
```diff
@@ -86,129 +89,30 @@
...
+    # prepare for battle
+    for knight in knights:
+        knight.prepare_to_battle()
 
-    # apply potion if exist
-    if mordred["potion"] is not None:
-        if "power" in mordred["potion"]["effect"]:
-            mordred["power"] += mordred["potion"]["effect"]["power"]
+    # Lancelot vs Mordred
+    knights[0] - knights[2]
```
vasylhnatiuk, line 107: i think it is not a good choice use `__sub__` to manage battle. it will be better create separate method fight and do it there

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #614, opened 2024-03-12
Solution

Vsevoloddeez:


vasylhnatiuk (changes requested):
fix flake8 tests

app/knight_stat.py
```diff
@@ -0,0 +1,20 @@
+def calculate_knight_stat(knight: dict) -> dict:
```
vasylhnatiuk, line 1: rewrite this function in Knight class as a method, and save data as attributes

vasylhnatiuk (changes requested):

PR #621, opened 2024-03-17
'Solution'

KravchenkoSerhiii:


app/KNIGHTS/knight.py
```diff
@@ -0,0 +1,22 @@
...
+    def potion_activate(self) -> None:
+        if self.potion is not None:
+            if "power" in self.potion["effect"]:
+                self.power += int(self.potion["effect"]["power"])
+
+            if "protection" in self.potion["effect"]:
+                self.protection += int(self.potion["effect"]["protection"])
+
+            if "hp" in self.potion["effect"]:
+                self.hp += int(self.potion["effect"]["hp"])
```
Dimosphen1, line 22: Do not use three if conditions, use a for loop and a combination of `setattr` and `getattr` methods

app/KNIGHTS/knights.py
Dimosphen1, line 1: Rename this file to `knights_config.py` to define it's purpose clearly. Besides that, do not use uppaer case in directory naming

app/KNIGHTS/knights.py
```diff
@@ -0,0 +1,86 @@
+Knights = {
```
Dimosphen1, line 1: In case it is a constant, it should be defined in the upper case

Dimosphen1 (changes requested):

PR #626, opened 2024-03-20
Solution

D-u-x:
- Create Knight class and implement methods for applying armour, weapon and potion stats
- Create Weapon class and init it's atributes
- Create Armour class and init it's atributes
- Create Potion class and init it's atributes
- Create knight_factory file that realising create_knight method
- Create camelot package that storing all files named above


EdAlekseiev (approved):
Good job

PR #627, opened 2024-03-20
Solution

Gilgumish:


app/main.py
```diff
@@ -85,130 +87,5 @@
...
-        lancelot["name"]: lancelot["hp"],
-        arthur["name"]: arthur["hp"],
-        mordred["name"]: mordred["hp"],
-        red_knight["name"]: red_knight["hp"],
-    }
-
-
-print(battle(KNIGHTS))
+results = battle(KNIGHTS)
+print(results)
```
DmytroHlazyrin, line 91: If you won't use the result - you don't have to write it to a variable

app/main.py
```diff
@@ -1,3 +1,5 @@
+from app.battles.battle import battle
+
 KNIGHTS = {
```
DmytroHlazyrin, line 3: in my opinion - it would be better to put configurations in a separate module

app/knights/knight.py
```diff
@@ -0,0 +1,34 @@
...
+            power: int,
+            hp: int,
+            armour: list,
+            weapon: dict,
+            potion: dict
+    ) -> None:
+        self.name = name
+        self.power = power
+        self.hp = hp
+        self.armour = [Armour(**a) for a in armour]
```
DmytroHlazyrin, line 17: Why did you use the short name "a" in the case of the armor? It is better to use a clearer name. 
It's the same in others.

DmytroHlazyrin (changes requested):

app/battles/battle.py
```diff
@@ -0,0 +1,14 @@
...
+    knight1.take_damage(knight2.power)
+    knight2.take_damage(knight1.power)
+
+
+def battle(knights: dict) -> dict:
+    knights = {name: Knight(**details) for name, details in knights.items()}
+    fight(knights["lancelot"], knights["mordred"])
+    fight(knights["arthur"], knights["red_knight"])
+
+    return {k.name: k.hp for k in knights.values()}
```
YuriiKindrat, line 14: what is k?)
use full names for vars

YuriiKindrat (changes requested):

Sofbog (approved):
LGTM

PR #629, opened 2024-03-20
first_try

Sydorovich:
Tried to implement unusable "name" parts of knight's equipment by adding some text to stdout that shows all equipment


NazarSheremet (approved):
Exelent solution!

PR #639, opened 2024-03-28
solution

SHCHERBANV:


app/battles/battle.py
```diff
@@ -0,0 +1,28 @@
...
+            if attacker["name"] != defender["name"]:
+                damage = max(0, attacker["power"] - defender["protection"])
+                defender["hp"] -= damage
+                defender["hp"] = max(0, defender["hp"])
+
+    return {knight["name"]: knight["hp"] for knight in knights_config.values()}
+
+
+KNIGHTS = {...}
+print(battle(KNIGHTS))
```
YuriiSmolii, line 28: What is the purpose of those lines?

YuriiSmolii (changes requested):
Your solution is seems to be not executable 
image

YuriiSmolii (approved):
LGTM

PR #640, opened 2024-03-28
Solution

fishray33:


YuriiSmolii (approved):
Could be better automated and should contain less hardcode.
In general acceptable.
Approved.

PR #646, opened 2024-04-04
Solution

sind14:


app/ready_to_battle.py
```diff
@@ -0,0 +1,24 @@
+def ready_to_battle(knights: dict) -> list[tuple]:
```
vasylhnatiuk, line 1: it is a too big function, try to divide this into a few function with different responsibilities

PR #648, opened 2024-04-06
Solution

SerhiiMar:


app/players/knight.py
```diff
@@ -0,0 +1,9 @@
+from app.players.player import Player
+
+
+class Knight(Player):
```
Oleksl888, line 4: what was the point of making an abstract class of player and then not expanding  it here? Why not simply create a Knight class instead of Player and write less code?

app/players/player.py
```diff
@@ -0,0 +1,88 @@
...
+    def unequip_armour_all(self) -> None:
+        for armour_part_name in self.inventory["armour"]:
+            self.unequip_armour(armour_part_name)
+
+    def prepare_for_battle(self) -> None:
+        self.equip_weapon()
+        self.equip_armour_all()
+        self.drink_potion()
+
+    def add_item_to_inventory(self) -> None:
```
Oleksl888, line 84: what is the point of this method?

app/items/equipment/eqipment.py
```diff
@@ -0,0 +1,14 @@
+from app.items.item import BaseItem
+
+
+class BaseEquipment(BaseItem):
+    def __init__(
```
Oleksl888, line 5: I would disagree with this design. The only thing that equipment is sharing is `name and is_equipped` attribute. With this every piece of equipment will have the same attrs which wouldnt make sense like hp/protection for  a sword.

Oleksl888 (changes requested):

app/items/item.py
```diff
@@ -0,0 +1,41 @@
...
+            item: Optional[dict]
+    ) -> Optional[Item]:
+        if not item:
+            return None
+        name = item.get("name") or item.get("part")
+        if item_class == "potion":
+            item = item["effect"]
+        power = item.get("power", 0)
+        hp = item.get("hp", 0)
+        protection = item.get("protection", 0)
```
LiudmylaKulinchenko, line 33: After fixes for this class you won't need creating this method. You can also use unpacking for this

LiudmylaKulinchenko (changes requested):

YuriiKindrat (approved):
GJ

PR #656, opened 2024-04-18
The solution of the task

Ja0ni:


app/knights/heroes.py
```diff
@@ -0,0 +1,36 @@
...
+            hp: int,
+            protection: int = 0
+    ) -> None:
+        self.name = name
+        self.power = power
+        self.hp = hp
+        self.protection = protection
+
+
+def init_hero(knights: dict) -> list[Hero]:
```
Dimosphen1, line 15: Use plural form since you return a list of elements
```suggestion
def init_heros(knights: dict) -> list[Hero]:
```

app/knights/heroes.py
```diff
@@ -0,0 +1,36 @@
+class Hero:
```
Dimosphen1, line 1: It would be better to use `Knight` class name so that it would be more relevant to the application context

app/main.py
```diff
@@ -1,214 +1,24 @@
...
+from app.knights.heroes import init_hero
+from app.knights.data import KNIGHTS
+
+
+def battle(knights_config: dict) -> dict:
+    heroes = init_hero(knights_config)
+    pairs = [
+        [
+            heroes[i], heroes[len(heroes) // 2 + i]
+        ] for i in range(len(heroes) // 2)
```
Dimosphen1, line 10: Do not use single-letter variables in the code

Dimosphen1 (changes requested):

vsmutok (approved):
Great!

PR #659, opened 2024-04-22
Done

Stekloduv:


app/main.py
```diff
@@ -1,213 +1,37 @@
...
-            arthur["power"] += arthur["potion"]["effect"]["power"]
+from app.Knights_attributes import KNIGHTS
 
-        if "protection" in arthur["potion"]["effect"]:
-            arthur["protection"] += arthur["potion"]["effect"]["protection"]
 
-        if "hp" in arthur["potion"]["effect"]:
-            arthur["hp"] += arthur["potion"]["effect"]["hp"]
+def prepare_knight(knight: dict) -> None:
+    knight["protection"] = sum(a["protection"] for a in knight["armour"])
```
Dimosphen1, line 5: Do not use single letter variable names in the code

app/main.py
```diff
@@ -1,213 +1,37 @@
...
+    knight["power"] += knight["weapon"]["power"]
+    if knight["potion"]:
+        for effect in ["power", "protection", "hp"]:
+            if effect in knight["potion"]["effect"]:
+                knight[effect] += knight["potion"]["effect"][effect]
```
Dimosphen1, line 10: Place this function to a separate file and import it here, there is no need to store it in the `app/main.py`

app/main.py
```diff
@@ -1,213 +1,37 @@
...
     red_knight["hp"] -= arthur["power"] - red_knight["protection"]
 
-    # check if someone fell in battle
-    if arthur["hp"] <= 0:
-        arthur["hp"] = 0
-
-    if red_knight["hp"] <= 0:
-        red_knight["hp"] = 0
+    for knight in [lancelot, mordred, arthur, red_knight]:
+        knight["hp"] = max(0, knight["hp"])
```
Dimosphen1, line 29: Place this logic in a separate function and place it in another file, so that yo will pass two knights instances and will receive their hp as a result of a fight.

app/Knights_attributes.py
Dimosphen1, line 1: Do not use upper case in the files name that contain python code

Dimosphen1 (changes requested):

vasylhnatiuk (approved):

PR #667, opened 2024-04-29
Solution

AlexNechval:


YuriiSmolii (approved):
LGTM but it could be done better with use of classes

PR #669, opened 2024-05-06
solution

RuslanLevchuk:


app/main.py
```diff
@@ -1,214 +1,29 @@
...
-    for a in lancelot["armour"]:
-        lancelot["protection"] += a["protection"]
+    print(Knight.knights["Lancelot"])
+    # run battles
+    Knight.battle(lancelot, mordred)
+    Knight.battle(red_knight, arthur)
 
-    # apply weapon
-    lancelot["power"] += lancelot["weapon"]["power"]
+    return {name: val.health for name, val in knights_dict.items()}
```
vsmutok, line 26: `val` not clear variable name

app/modules/weapon.py
```diff
@@ -0,0 +1,16 @@
+from __future__ import annotations
```
vsmutok, line 1: I think this import redundant here

app/modules/knight.py
```diff
@@ -0,0 +1,66 @@
...
+    def apply_features(other: Knight) -> None:
+        other.power += other.weapon.damage_power + sum(
+            [0 if not other.potion else other.potion.effect["power"]]
+        )
+        other.health += sum(
+            [0 if not other.potion else other.potion.effect["hp"]]
+        )
+        other.protection = sum(other.armour) + sum(
+            [0 if not other.potion else other.potion.effect["protection"]]
+        )
```
vsmutok, line 44: DRY 

think about how it can be optimized so that there is less duplication of code

app/main.py
```diff
@@ -1,214 +1,29 @@
...
+from app.modules.knights import knights as knights_origin_dict
 
 
-def battle(knightsConfig):
-    # BATTLE PREPARATIONS:
+def battle(knights: dict) -> dict:
+    # storage of all knights objects in a dictonary
+    knights_dict = {}
+    # parse inner knights dict. In an each iteration adds all knight features
+    for kn in knights.values():
```
vsmutok, line 9: `kn` not clear variable name

vsmutok (changes requested):

app/main.py
```diff
@@ -1,214 +1,36 @@
...
+        knights_dict[knight_unit["name"]].apply_weapon(knight_unit["weapon"])
+        knights_dict[knight_unit["name"]].apply_potion(knight_unit["potion"])
+
+    # create instance variables for better reading of code
+    lancelot = Knight.knights["Lancelot"]
+    arthur = Knight.knights["Arthur"]
+    mordred = Knight.knights["Mordred"]
+    red_knight = Knight.knights["Red Knight"]
+
+    print(Knight.knights["Lancelot"])
```
vasylhnatiuk, line 25: why do you need this? dlete useless print()

app/modules/knight.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def apply_potion(self, potion: dict) -> None:
+        if not potion:
+            return None
+        self.potion = Potion(potion["name"], potion["effect"])
+
+    # apply all features if a knight before fight
+    @staticmethod
+    def apply_features(other: Knight) -> None:
+        def parser(potion_instance: Potion | None, effect_key: str) -> list:
```
vasylhnatiuk, line 36: don't write function in another function if it is not a decorator

app/modules/knight.py
```diff
@@ -0,0 +1,72 @@
...
+            return [
+                0 if not potion_instance
+                else potion_instance.effect[effect_key]
+            ]
+
+        other.power += other.weapon.damage_power + sum(
```
vasylhnatiuk, line 42: instead of duplicating use loop and settatr

vasylhnatiuk (changes requested):

app/main.py
RuslanLevchuk: Because original code uses printing of lancelot info

vasylhnatiuk (approved):

PR #672, opened 2024-05-12
Solution

AlexTarasov57:


app/main.py
```diff
@@ -86,122 +88,30 @@
...
-            red_knight["protection"] += red_knight["potion"]["effect"]["protection"]
-
-        if "hp" in red_knight["potion"]["effect"]:
-            red_knight["hp"] += red_knight["potion"]["effect"]["hp"]
+    red_knight = Knights.characteristics(
+        knights_configs=knights_config["red_knight"])
 
     # -------------------------------------------------------------------------------
     # BATTLE:
-
```
Dimosphen1, line 182: Remove redundant comments

app/main.py
```diff
@@ -1,3 +1,5 @@
+from app.characteristics_of_a_knight.knight import Knights
+from app.Battle.battle_knight import BattleKnights
 KNIGHTS = {
```
Dimosphen1, line 3: Move `KNIGHTS` variable to a separate file `knights_config.py` to not overload `main.py` file

app/characteristics_of_a_knight/knight.py
```diff
@@ -0,0 +1,25 @@
+class Knights:
+    @staticmethod
+    def characteristics(knights_configs: dict) -> dict:
+        # apply armour
```
Dimosphen1, line 4: There is no need to add this comments, code is self-descriptive enough

app/characteristics_of_a_knight/knight.py
```diff
@@ -0,0 +1,25 @@
+class Knights:
+    @staticmethod
+    def characteristics(knights_configs: dict) -> dict:
```
Dimosphen1, line 3: In case a class consists of only one static method, you could create it as a separate function and not create a class for it. Besides that, function names should describe actions in the form of a verb. The same comment with the static method relates to the `BattleKnights` class as well.

app/Battle/battle_knight.py
```diff
@@ -0,0 +1,12 @@
+class BattleKnights:
+    @staticmethod
+    def battle_all(first: dict, second: dict) -> None:
+        first["hp"] -= second["power"] - first["protection"]
+        second["hp"] -= first["power"] - second["protection"]
+
+        # check if someone fell in battle
```
Dimosphen1, line 7: Remove redundant comment

Dimosphen1 (changes requested):

PR #673, opened 2024-05-14
add solution

midetc:


app/models/knight.py
```diff
@@ -0,0 +1,31 @@
...
+            weapon: dict,
+            potion: None | dict
+    ) -> None:
+        self.name = name
+        self.hp = hp
+        self.power = power + weapon["power"]
+        self.protection = sum([i["protection"] for i in armour])
+
+        if potion is not None:
+            potion_effect = potion["effect"]
```
vasylhnatiuk, line 20: instead of duplicating use loop and `settatr()`

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #679, opened 2024-05-19
Solution

SuskyiVolodymyr:


app/battle.py
```diff
@@ -0,0 +1,16 @@
...
+    knight1.hp = attack(knight1.hp, knight2.power, knight1.protection)
+    knight2.hp = attack(knight2.hp, knight1.power, knight2.protection)
+
+
+def attack(hp: int, power: int, protection: int) -> int:
+    return is_hp_lt_zero(hp - (power - protection))
+
+
+def is_hp_lt_zero(hp: int) -> int:
+    if hp < 0:
```
DevSpaciX, line 14: You can do it in one line ))

DevSpaciX (changes requested):
All looks good , but a little optimization requested : )

app/battle.py
Oleg-s-s, line 1: These functions are all related to Knights, in my opinion it would be better to make static methods and use them, but at least you separated this logic from others. Just consider if your function is only working with one class and its attributes, that means that using `staticmethod` will be the best approach.

Oleg-s-s (approved):

PR #682, opened 2024-05-26
Solution

lilarin:


app/knight.py
```diff
@@ -0,0 +1,42 @@
...
+    def apply_potion(self) -> None:
+        if self.potion:
+            potion = self.potion.get("effect", {})
+            for effect, value in potion.items():
+                if effect == "power":
+                    self.power += value
+                elif effect == "protection":
+                    self.protection += value
+                elif effect == "hp":
+                    self.hp += value
```
Oleg-s-s, line 37: Its ok, but I dont like how you are increasing levels.
1st thing that you can do is simply do it reversed.
```suggestion
        if not self.potion:
            return

        potion = self.potion.get("effect", {})
        for effect, value in potion.items():
            if effect == "power":
                self.power += value
            elif effect == "protection":
                self.protection += value
            elif effect == "hp":
                self.hp += value
```
And if we think about optimizing I would use mapping to simplify adding new effects
```suggestion
def apply_potion(self) -> None:
    if not self.potion:
        return
    
    effect_map = {
        "power": "power",
        "protection": "protection",
        "hp": "hp"
    }

    for effect, value in self.potion.get("effect", {}).items():
        if effect in effect_map:
            setattr(self, effect_map[effect],
                    getattr(self, effect_map[effect]) + value)
```

Oleg-s-s (approved):

PR #683, opened 2024-05-30
solution

AndreiVed:


vasylhnatiuk (changes requested):

app/character/knight.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def apply_armour(self, armour: dict) -> None:
+        if armour is not None:
+            for ar in armour:
+                self.protection += ar["protection"]
+
+    def apply_potion(self, potion: dict) -> None:
+        if potion is not None:
+            potion_effect = potion["effect"]
+            if "power" in potion_effect:
```
vasylhnatiuk, line 48: instead of duplicating use settatr()

app/character/duel.py
```diff
@@ -0,0 +1,11 @@
+def duel(knight1: object, knight2: object) -> dict:
```
vasylhnatiuk, line 1:
```suggestion
def duel(knight1: Knight, knight2: Knight) -> dict:
```

vasylhnatiuk (changes requested):

app/character/knight.py
vasylhnatiuk: do it - `instead of duplicating use settatr()`

AndreiVed: I don't understand how I can do it by setattr() :((

vasylhnatiuk (changes requested):
you pushed nothing

AndreiVed:
@vasylhnatiuk  please, help me with it
Screenshot 2024-05-30 at 16 33 55

vasylhnatiuk (approved):
good job

PR #684, opened 2024-05-31
Solution

pikaOpika:


app/main.py
```diff
@@ -1,213 +1,20 @@
...
-    if red_knight["hp"] <= 0:
-        red_knight["hp"] = 0
-
-    # Return battle results:
+from app.people.KNIGHTS import KNIGHTS
+from app.transform import transform_knights
+
+
+def battle(knights: dict) -> dict:
+    list_of_knights = transform_knights(knights)
```
pavlopro, line 6: You could create a dictionary, e.g., knights_battle = {"lancelot": "mordred", "arthur": "red_knight"}.

It would be better if this method returned a dictionary instead of a list, something like this: {"lancelot": object of lancelot, ...}. Then you could iterate over it and call the equip method.

Then you could do something like this:
knights_battle["lancelot"].fight(knights_battle["mordred"])

pavlopro (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #690, opened 2024-06-19
Solution

Serveladik:


NazarSheremet (changes requested):
You can't delete test case from `tests/test_main.py`. Return it

Serveladik:
need a check

PR #704, opened 2024-07-05
Solution

tylerj231:


app/knights/knight.py
```diff
@@ -0,0 +1,57 @@
...
+        if self.potion is not None:
+            effect = self.potion["effect"]
+            if "power" in effect:
+                self.power += effect["power"]
+
+            if "protection" in effect:
+                self.protection += effect["protection"]
+
+            if "hp" in effect:
+                self.hp += effect["hp"]
```
EdAlekseiev, line 43: instead of many if try to use loop and settatr()

EdAlekseiev (changes requested):

YukihiroSM (approved):
Looks good to me!

PR #709, opened 2024-07-11
Completed all tasks (and also isn't short description below)

dimak20:
First of all, this task is the most interesting thing I've ever done so far. I tried implementing end code with maximum depersonalization - in case when "knight"s' names would change. Frankly, that would be even more short, if I implemented all characteristics right in the Knight's instance without separating by weapon, armor etc. But it would be less flexible - for instance, when user or someone want to change weapon and Knight's instance attribute (weapon.name in dict if dict was as like attribute) would conflict with new weapon.name. The worst thing I faced and struggled with while I was creating a solution - circluar import. That's why I needed to import class inside the function... Of course, I might have implemented all "knight's equipment" inside just one class or file, but that would not seemed good to me.

### DIFF
--- a/app/battle_logic/fight.py
+++ b/app/battle_logic/fight.py
@@ -1,0 +1,15 @@
+from app.battle_preparation.knight import Knight
+
+
+class Fight:
+    def __init__(self, first_knight: Knight, second_knight: Knight) -> None:
+        self.first = first_knight
+        self.second = second_knight
+
+    def begin_battle(self) -> None:
+        self.first.hp -= self.second.power - self.first.protection
+        self.second.hp -= self.first.power - self.second.protection
+        if self.first.hp < 0:
+            self.first.hp = 0
+        if self.second.hp < 0:
+            self.second.hp = 0
--- a/app/battle_preparation/armour.py
+++ b/app/battle_preparation/armour.py
@@ -1,0 +1,20 @@
+from app.battle_preparation.knight import Knight
+
+
+class Armour:
+    def __init__(self, equipment: list, protection: int) -> None:
+        self.equipment = equipment
+        self.protection = protection
+
+    @classmethod
+    def prepare_armour(cls, parts: list) -> "Armour":
+        protection = 0
+        equipment = []
+        for part in parts:
+            equipment.append(part["part"])
+            protection += part["protection"]
+        return cls(equipment, protection)
+
+    def consider_armour_effect(self, knight: Knight) -> None:
+        knight.protection = self.protection
+        knight.equipment = self.equipment
--- a/app/battle_preparation/knight.py
+++ b/app/battle_preparation/knight.py
@@ -1,0 +1,24 @@
+class Knight:
+    def __init__(self, name: str, power: int, hp: int) -> None:
+        self.name = name
+        self.power = power
+        self.hp = hp
+        self.protection = 0
+        self.equipment = None
+        self.liquid = None
+        self.weapon = None
+
+    def get_ready_for_battle(self,
+                             armor_parts: list,
+                             miracle_potion: dict,
+                             divine_weapon: dict) -> None:
+        from app.battle_preparation.armour import Armour
+        armour = Armour.prepare_armour(armor_parts)
+        armour.consider_armour_effect(self)
+        from app.battle_preparation.potion import Potion
+        potion = Potion.drink_potion(miracle_potion)
+        if potion:
+            potion.consider_potion_effect(self)
+        from app.battle_preparation.weapon import Weapon
+        weapon = Weapon.prepare_weapon(divine_weapon)
+        weapon.consider_weapon(self)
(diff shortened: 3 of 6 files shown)


app/main.py
```diff
@@ -86,128 +90,23 @@
...
+def battle(knights_config: dict) -> dict:
+    participants = []
+    for name, info in knights_config.items():
+        participant = Knight(info["name"],
+                             info["power"],
+                             info["hp"])
+        participants.append(participant)
+        participant.get_ready_for_battle(info["armour"],
+                                         info["potion"],
+                                         info["weapon"])
```
MNDonut, line 102: Read please this message
https://mate.academy/chats/244801?selectedMessageId=3129519

app/battle_preparation/knight.py
```diff
@@ -0,0 +1,24 @@
...
+        from app.battle_preparation.armour import Armour
+        armour = Armour.prepare_armour(armor_parts)
+        armour.consider_armour_effect(self)
+        from app.battle_preparation.potion import Potion
+        potion = Potion.drink_potion(miracle_potion)
+        if potion:
+            potion.consider_potion_effect(self)
+        from app.battle_preparation.weapon import Weapon
+        weapon = Weapon.prepare_weapon(divine_weapon)
+        weapon.consider_weapon(self)
```
MNDonut, line 24: if you import in this way, then all imports should be at the top of the function

app/battle_preparation/armour.py
```diff
@@ -0,0 +1,20 @@
+from app.battle_preparation.knight import Knight
+
+
+class Armour:
+    def __init__(self, equipment: list, protection: int) -> None:
+        self.equipment = equipment
+        self.protection = protection
+
+    @classmethod
+    def prepare_armour(cls, parts: list) -> "Armour":
```
MNDonut, line 10: don't use quotes, you can import annotations from future

app/battle_logic/fight.py
```diff
@@ -0,0 +1,15 @@
...
+        self.first = first_knight
+        self.second = second_knight
+
+    def begin_battle(self) -> None:
+        self.first.hp -= self.second.power - self.first.protection
+        self.second.hp -= self.first.power - self.second.protection
+        if self.first.hp < 0:
+            self.first.hp = 0
+        if self.second.hp < 0:
+            self.second.hp = 0
```
MNDonut, line 15: ternary operator can be used here

app/battle_preparation/knight.py
```diff
@@ -0,0 +1,24 @@
+class Knight:
+    def __init__(self, name: str, power: int, hp: int) -> None:
+        self.name = name
+        self.power = power
+        self.hp = hp
+        self.protection = 0
+        self.equipment = None
```
MNDonut, line 7: do you set this attribute anywhere?

app/battle_preparation/knight.py
MNDonut, line 24: I a bit disagree with this approach. You create class instances in order to just call `consider` method and for attributes self.protection, equipment, liquid, weapon you save only values(numbers and for potion it's potion name for some reason). So you actually don't need class instances, just some static methods. But from the other side you don't save class instances and it means if you leave battle function, you'll not be able to access what list of armours this knight has and will need to extract it again from config. I recommend to save this boosts as class instance / list of instances and have some methods/properties in Knight that will call methods of these instances(weapon, armour, etc) that will return boost(for example number of protection)
In case you want to get info about what armour this knight has, you'll be able to get it using knight.armour -> list[Armour] and calculate that boost whenever you want

MNDonut (changes requested):

app/battle_preparation/knight.py
dimak20: If I import this at the top of the function, It will cause circle import error. Thus, I need to refactor this modules

app/battle_preparation/knight.py
dimak20: Yes, in the main module. Protection for calculating armour's effect and equipment just for info

app/battle_preparation/knight.py
dimak20: Just wanted to separate this modules in order to make this boosts more flexible

dimak20:
I fixed imports and etc., but remained classes and two methods for each part of knignt's equipment and have some reasons for this:
1) It might happen that, for instance, we have some buffs and all potions would have x2 effect and that would be easy just to implement in Potion class.
2)Certain knight might have debuff and considering potion or other part of knight's equipment would not be completely counted as we can see in his list attribute. So, "considering" methods can help implement this case

MNDonut (changes requested):
Yes, everything you mentioned is correct. Classes should be definitely used here. What I meant is that you don't have any relationship between knight and armour for example. Armour is independent in you case, but actually it belongs to knight as well. You have got the config where knight has armour, weapon, potion, but if you need to print knight's armour, you won't be able to do it. I recommend you to make some relationship between knight and boosts and calculate value of each boost using class methods

MNDonut (approved):

PR #716, opened 2024-07-24
Solution

zazmarga:


app/participants/knights.py
```diff
@@ -0,0 +1,41 @@
...
+    def apply_armour(self, armour: list) -> None:
+        for one in armour:
+            self.protection += one["protection"]
+
+    def apply_weapon(self, weapon: dict) -> None:
+        self.power += weapon["power"]
+
+    def apply_potion(self, potion: dict) -> None:
+        for one in potion:
+            if one == "power":
```
vasylhnatiuk, line 27: DRY, use loop and settatr()

app/participants/opponents.py
```diff
@@ -0,0 +1,7 @@
+def appoint_opponent(participants_knights: dict) -> None:
```
vasylhnatiuk, line 1: it doesn't make sense create a file for only one function

vasylhnatiuk (changes requested):
it doesn't make sense create file for one function

vasylhnatiuk (approved):

PR #718, opened 2024-07-26
create basic classes, functions

vladislav-tsybuliak1:
characters:
  -> **knights**: they can _wear_armour_, _take_weapon_, _drink_potion_, _fight_with_ and _check_hp_
  
gears:
  -> **armours**
  -> **weapons**
  -> **potions**
 Each gear has a method to create gear instances from list/dicts

actions:
  -> **prepare_characters**: a function that prepares characters for battle
  -> **fight**: functions that puts two characters in a fight
  -> **show_battle_results**: return results of the battle


app/characters/knights.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def drink_potion(self, potion: Potion) -> None:
+        if potion:
+            print(f"{self.name} drank {potion.name}.")
+            if "power" in potion.effect:
+                self.power += potion.effect["power"]
+            if "hp" in potion.effect:
+                self.hp += potion.effect["hp"]
+            if "protection" in potion.effect:
+                self.protection += potion.effect["protection"]
```
MNDonut, line 43: let's use vice-versa condition in order to move other ifs out of this if

app/characters/knights.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def fight_with(self, other: Knight) -> None:
+        self.hp -= other.power - self.protection
+        other.hp -= self.power - other.protection
+
+    def check_hp(self) -> None:
+        if self.hp < 0:
```
MNDonut, line 50: when hp == 0 the knight is dead as well

MNDonut (changes requested):

MNDonut (approved):

PR #720, opened 2024-07-26
Test branch

ihorutkin:


app/main.py
```diff
@@ -1,213 +1,58 @@
...
+            armour=value["armour"],
+            weapon=value["weapon"],
+            potion=value["potion"]
+        )
+        knight.get_ready()
+        knights_list.append(knight)
+    lancelot = knights_list[0]
+    arthur = knights_list[1]
+    mordred = knights_list[2]
+    red_knight = knights_list[3]
```
MNDonut, line 23: you can use unpacking here

app/main.py
```diff
@@ -1,213 +1,58 @@
...
     # check if someone fell in battle
-    if arthur["hp"] <= 0:
-        arthur["hp"] = 0
+    if arthur.hp <= 0:
+        arthur.hp = 0
 
-    if red_knight["hp"] <= 0:
-        red_knight["hp"] = 0
+    if red_knight.hp <= 0:
+        red_knight.hp = 0
```
MNDonut, line 48: you can use ternary operator

app/characters/knight.py
```diff
@@ -0,0 +1,46 @@
+from __future__ import annotations
+
+
+class Knight:
+    def __init__(
+        self,
+        name: str,
+        power: int,
+        hp: int,
+        armour: list,
```
MNDonut, line 10: knight may not have armour

app/characters/knight.py
```diff
@@ -0,0 +1,46 @@
...
+        self.hp = hp
+        self.armour = armour
+        self.weapon = weapon
+        self.potion = potion
+        self.protection = protection
+
+    def get_armour(self) -> None:
+        self.protection = 0
+        for elem in self.armour:
+            self.protection += elem["protection"]
```
MNDonut, line 26: let's do it in 1 line

app/characters/knight.py
```diff
@@ -0,0 +1,46 @@
...
+    def get_potion(self) -> None:
+        if self.potion is not None:
+            if "power" in self.potion["effect"]:
+                self.power += self.potion["effect"]["power"]
+
+            if "protection" in self.potion["effect"]:
+                self.protection += self.potion["effect"]["protection"]
+
+            if "hp" in self.potion["effect"]:
+                self.hp += self.potion["effect"]["hp"]
```
MNDonut, line 40: Let's simplify nested conditions using reverse condition
```
if not self.potion:
     return

# other if statements, but not nested inside another if statement
```

MNDonut, line 40: let's save effect in variable and reuse it + use .get when fetch kind of potion

MNDonut (changes requested):

app/characters/knight.py
```diff
@@ -21,23 +21,18 @@ def __init__(
         self.protection = protection
 
     def get_armour(self) -> None:
-        self.protection = 0
-        for elem in self.armour:
-            self.protection += elem["protection"]
+        self.protection = sum([elem["protection"] for elem in self.armour])
```
MNDonut, line 24: [] are redundant, sum works with generator as well

app/characters/knight.py
```diff
@@ -21,23 +21,18 @@ def __init__(
...
-
-            if "hp" in self.potion["effect"]:
-                self.hp += self.potion["effect"]["hp"]
+    def get_potion(self) -> int | None:
+        if self.potion is None:
+            return 0
+        effect = self.potion.get("effect", {})
+        self.power += effect.get("power", 0)
+        self.protection += effect.get("protection", 0)
+        self.hp += effect.get("hp", 0)
```
MNDonut, line 35: looks much more better ;)
You can read about Cognitive complexity more(it's the issue you had here)

app/main.py
```diff
@@ -17,10 +17,13 @@ def battle(knights_config: dict) -> dict:
...
-    arthur = knights_list[1]
-    mordred = knights_list[2]
-    red_knight = knights_list[3]
+    (lancelot,
+     arthur,
+     mordred,
+     red_knight) = (knights_list[0],
+                    knights_list[1],
+                    knights_list[2],
+                    knights_list[3])
```
MNDonut, line 26: It isn't unpacking actually ;)
You can do it in this way
```
lancelot, arthur, mordred, red_knight = knights_list
```
If there are enough values, python will distribute values among knights, otherwise error will be raised

MNDonut (changes requested):

app/main.py
```diff
@@ -17,13 +17,13 @@ def battle(knights_config: dict) -> dict:
...
-                    knights_list[1],
-                    knights_list[2],
-                    knights_list[3])
+    knights = [
+        knights_list[0],
+        knights_list[1],
+        knights_list[2],
+        knights_list[3]
+    ]
+    lancelot, arthur, mordred, red_knight = knights
```
MNDonut, line 26: Noo, the idea is to not use indexes at all. You have a list of knights. You know their order. Thus you can unpack it in 1 line as I mentioned before

MNDonut (changes requested):

app/main.py
ihorutkin: Yeah, got you, it is done

MNDonut (approved):
Perfect

PR #721, opened 2024-07-27
Develop

Narberal90:


Narberal90:
Tried to make good imports, but "Knight" import in "apply" methods causes circular import and error

app/human/knight.py
```diff
@@ -0,0 +1,19 @@
...
+        self.power = power
+        self.hp = hp
+        self.armor = armor
+        self.equip_list = None
+        self.weapon = None
+        self.config = None
+
+    def prepare_for_battle(self, config: dict) -> None:
+        self.config = config
+        for cl in (Armor, Weapon, Potion):
```
MNDonut, line 18: I would create a class constant, called something like "UPGRADES" and move these classes to this constant so that it's easier to manage these boosts, but in general idea is awesome

app/equipment/armor.py
```diff
@@ -0,0 +1,24 @@
+from typing import Any
+
+from app.equipment.equipment import Equipment
+
+
+class Armor(Equipment):
+    def __init__(self, part: str, power: int) -> None:
+        super().__init__(part)
+        self.power = power
```
MNDonut, line 9: Armour should give protection, not power
image

app/equipment/armor.py
```diff
@@ -0,0 +1,24 @@
...
+        super().__init__(part)
+        self.power = power
+
+    @classmethod
+    def apply(cls, knight: Any, source: dict) -> None:
+        armor_data = (
+            source.get(knight.name.lower(),
+                       source.get(knight.name.replace(" ", "_").lower(), {})
+                       ).get("armour", [])
+        )
```
MNDonut, line 17: it should be either good formatted or use variables here

app/equipment/weapon.py
```diff
@@ -0,0 +1,28 @@
...
+    @classmethod
+    def apply(cls, knight: Any, source: dict) -> None:
+        weapon_data = (
+            source.get(
+                knight.name.lower(),
+                source.get(
+                    knight.name.replace(" ", "_").lower(), {}
+                )
+            ).get("weapon", None)
+        )
```
MNDonut, line 20: What do you do here?
You loop through all knights and their configs here
image

This is weapon dict of specific knight. So you just need to fetch power and add it to knight's power, ain't I right?
```
"weapon": {
        "name": "Two-handed Sword",
        "power": 55,
},
```

app/equipment/weapon.py
```diff
@@ -0,0 +1,28 @@
...
+        if weapon_data:
+            knight.weapon = (
```
MNDonut, line 22: weapon data is required
image

app/equipment/armor.py
MNDonut: The same that with weapon. You can just fetch using "armour"

app/equipment/potion.py
```diff
@@ -0,0 +1,26 @@
...
+        super().__init__(name)
+        self.effect = effect
+
+    @classmethod
+    def apply(cls, knight: Any, source: dict) -> None:
+        potion_data = (
+            source.get(knight.name.lower(),
+                       source.get(knight.name.replace(" ", "_").lower(), {})
+                       ).get("potion", None)
+        )
```
MNDonut, line 17: just fetch potion?

app/main.py
```diff
@@ -86,129 +88,33 @@
...
-    return {
-        lancelot["name"]: lancelot["hp"],
-        arthur["name"]: arthur["hp"],
-        mordred["name"]: mordred["hp"],
-        red_knight["name"]: red_knight["hp"],
-    }
+        if knight1.hp <= 0:
+            knight1.hp = 0
+        if knight2.hp <= 0:
+            knight2.hp = 0
```
MNDonut, line 115: you can use ternary operator here

MNDonut (changes requested):

Narberal90:
- Added a constant in the Knight class.
- Improved and simplified the logic in the Weapon, Armor, and Potion classes.
- Also I changed the "armor" attribute of the "Knight" class to "protection" 
  so as not to confuse armor as a class and as armor points

app/equipment/armor.py
```diff
@@ -4,21 +4,15 @@
...
     def apply(cls, knight: Any, source: dict) -> None:
-        armor_data = (
-            source.get(knight.name.lower(),
-                       source.get(knight.name.replace(" ", "_").lower(), {})
-                       ).get("armour", [])
-        )
-
-        if armor_data:
+        if source.get("armour"):
             knight.equip_list = []
```
MNDonut, line 14: Btw, why to not set equip_list = [] in Knight.__init__ instead of this place? It seems to me that you don't use some checks for None of this attribute

app/equipment/armor.py
```diff
@@ -4,21 +4,15 @@
...
     @classmethod
     def apply(cls, knight: Any, source: dict) -> None:
-        armor_data = (
-            source.get(knight.name.lower(),
-                       source.get(knight.name.replace(" ", "_").lower(), {})
-                       ).get("armour", [])
-        )
-
-        if armor_data:
+        if source.get("armour"):
```
MNDonut, line 13: In general we can save it in variable and reuse below

app/equipment/armor.py
```diff
@@ -4,21 +4,15 @@
...
             knight.equip_list = []
-            for armor in armor_data:
-                armor = Armor(part=armor["part"], power=armor["protection"])
+            for armor in source.get("armour"):
+                armor = Armor(armor["part"], armor["protection"])
```
MNDonut, line 16: Invite you to attend today's QnA at 14:00, will show you some trick ;)
Or you can use this tip: you can use unpacking here to make it look better

app/main.py
```diff
@@ -90,10 +90,11 @@
 
 def battle(knights_config: dict) -> None:
     knights = {}
-    for name, config in knights_config.items():
+    for config in knights_config.values():
         knight = Knight(config["name"], config["power"], config["hp"])
-        knight.prepare_for_battle(knights_config)
+        knight.prepare_for_battle(config)
         knights[knight.name] = knight
+        print(knight.__dict__)
```
MNDonut: you can remove it

app/equipment/potion.py
```diff
@@ -0,0 +1,21 @@
...
+    def apply(cls, knight: Any, source: dict) -> None:
+        if source.get("potion"):
+            potion_effect = source["potion"]["effect"]
+            for characteristic_name, value in potion_effect.items():
+                if characteristic_name == "power":
+                    knight.power += value
+                elif characteristic_name == "hp":
+                    knight.hp += value
+                elif characteristic_name == "protection":
+                    knight.protection += value
```
MNDonut, line 21: When you do so many nested ifs and loops, it's called Cognitive complexity and we should reduce it
Let's use vice-versa condition of get potion, in this way the loop will be moved outside

MNDonut (changes requested):

MNDonut (approved):

PR #722, opened 2024-07-27
Solution

Bogdan-Rylov:


app/main.py
```diff
@@ -1,3 +1,5 @@
+from app.knight import Knight
+
```
MNDonut, line 2: 2 blank lines should be added after the last import

app/knight.py
```diff
@@ -0,0 +1,65 @@
...
+        self.base_hp = hp
+        self.hp = hp
+        self.armor = armor
+        self.weapon = weapon
+        self.potion = potion
+        self.protection = 0
+        self.apply_gear_and_items()
+
+    @staticmethod
+    def parse_from_dict(knight_dict: dict) -> "Knight":
```
MNDonut, line 28: don't use quotes. if you get error then import annotations from future

app/knight.py
```diff
@@ -0,0 +1,65 @@
...
+        )
+        potion = (
+            Potion(
+                knight_dict["potion"]["name"],
+                knight_dict["potion"]["effect"]
+            )
+            if knight_dict["potion"]
+            else None
+        )
+        return Knight(name, power, hp, armor, weapon, potion)
```
MNDonut, line 48: basically classmethod is more appropriate here as you return class instance of the same class

app/knight.py
```diff
@@ -0,0 +1,65 @@
...
+            self,
+            name: str,
+            power: int,
+            hp: int,
+            armor: list[ArmorPart],
+            weapon: Weapon,
+            potion: Potion
+    ) -> None:
+        self.name = name
+        self.base_power = power
```
MNDonut, line 17: do you use it somewhere?

app/knight.py
```diff
@@ -0,0 +1,65 @@
...
+                knight_dict["potion"]["name"],
+                knight_dict["potion"]["effect"]
+            )
+            if knight_dict["potion"]
+            else None
+        )
+        return Knight(name, power, hp, armor, weapon, potion)
+
+    def apply_gear_and_items(self) -> None:
+        self.hp = self.base_hp
```
MNDonut, line 51: From what I see in init, self.hp and self.base hp are the same, why do you assign it here?
Also I think we can get rid of self.base_hp

MNDonut (changes requested):

app/knight.py
Bogdan-Rylov: I added base_hp and base_power, so a knight could fight several times (his hp and power may change after fight / drinking a potion). Those attributes enable restoring the hp and power to default values

app/main.py
Bogdan-Rylov: For some reason, Ctrl + Alt + L leaves 1 empty space here

MNDonut (approved):
Regarding CTRL + ALT + L in pycharm(that you mentioned)
It just reformats the code, split's imports, etc, but doesn't follow PEP8 ;)

PEP8 doc about blank lines and imports
https://peps.python.org/pep-0008/#blank-lines

More details:
https://stackoverflow.com/questions/2953250/python-pep8-blank-lines-convention

What actually this combination does:
https://www.jetbrains.com/help/pycharm/reformat-file-dialog.html#:~:text=Reformat%20File%20dialog%EF%BB%BF&text=If%20you%20choose%20Code%20%7C%20Reformat,to%20Reformat%20and%20rearrange%20code.

PR #723, opened 2024-07-31
Solution

DSahalatyi:
__

### DIFF
--- a/app/knights/knight.py
+++ b/app/knights/knight.py
@@ -1,0 +1,41 @@
+from app.knights.inventory.weapon import Weapon
+from app.knights.inventory.armour import Armour
+from app.knights.inventory.consumables import Potion
+
+
+class Knight:
+    def __init__(self, knight):
+        self.name = knight["name"]
+        self.power = knight["power"]
+        self.hp = knight["hp"]
+        self.protection = 0
+        self.weapon = Weapon(knight["weapon"])
+        self.armour = [
+            Armour(part["part"], part["protection"])
+            for part in knight["armour"]
+        ]
+        self.potion = Potion.add(knight["potion"])
+
+    def __str__(self):
+        return self.name
+
+    def get_ready(self):
+        self.apply_armour()
+        self.apply_weapon()
+        if self.potion:
+            self.use_potion()
+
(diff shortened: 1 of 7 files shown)


app/knights/knight.py
```diff
@@ -0,0 +1,43 @@
+from __future__ import annotations
+
+from app.knights.inventory.weapon import Weapon
+from app.knights.inventory.armour import Armour
+from app.knights.inventory.consumables import Potion
+
+
+class Knight:
+    def __init__(self, knight: dict) -> None:
```
Oleksl888, line 9: Here as well would be better to have many params instead of `knight: dict` - this way you make it obvious which params are mandatory and which ones are optional.

app/knights/knight.py
```diff
@@ -0,0 +1,43 @@
...
+        self.name = knight["name"]
+        self.power = knight["power"]
+        self.hp = knight["hp"]
+        self.protection = 0
+        self.weapon = Weapon(knight["weapon"])
```
Oleksl888, line 14: Creating instances inside of `__init__` is possible but it makes logic of the class quite complicated. Is a good practice to break down complex logic into smaller pieces, like here you could create instances of inventory classes elsewhere and simply pass created instances to `__init__`
```
class Knight:

    def __init__(
            self,
            name: str,
            power: int,
            hp: int,
            weapon: Weapon,
            armour: Optional[list[Armour]] = None,
            potion: Optional[Potion] = None,
    ) -> None:
        self.name = name
        self.power = power
        self.hp = hp
        self.protection = 0
        self.armour = armour or []
        self.weapon = weapon
        self.potion = potion

Oleksl888 (approved):
I really like what you did. This is the best implementation of this task that I have seen so far. I like how the code is structured - good job on creating packages based on what the classes do. The modules are also well structured, it is good to break code in smaller chunks instead of keeping most logic in one file. 
I left few comments for your consideration - it doesn't mean you did it wrong, I just think it can be more explicit when defining parameters in constructor. But this task does not have strict guidelines, it is mostly about you showing of your skills and understanding how structure your programs.
Good job!

PR #728, opened 2024-08-03
solution

BastovOleksandr:
I had to solve the problem a second time because I optimized not only the code itself, but also the input dictionary, and the tests were written according to it, although the task said that you can do whatever you want, as long as it works :(
In the first solution, I had several directories and many classes (most without implementation, but as a clear example of what the game could be and that it could expand)

### DIFF
--- a/app/hero.py
+++ b/app/hero.py
@@ -1,0 +1,94 @@
+from __future__ import annotations
+
+
+class Hero:
+    def __init__(
+            self,
+            name: str = "Default knight",
+            power: int = 20,
+            hp: int = 50,
+            armor: list = None,
+            weapon: dict = None,
+            potion: dict = None
+    ) -> None:
+        self.name = name
+        self.power = power
+        self.hp = hp
+        self.armor = armor if armor else []
+        self.weapon = weapon if weapon else {}
+        self.potion = potion if potion else {}
+        self.protection = 0
+
+    @classmethod
+    def create_from_config(cls, config: dict) -> Hero:
+        return cls(
+            name=config.get("name"),
+            power=config.get("power"),
+            hp=config.get("hp"),
+            armor=config.get("armour"),
+            weapon=config.get("weapon"),
+            potion=config.get("potion")
+        )
+
+    def put_on_armor(self) -> None:
+        if self.armor:
+            for item in self.armor:
+                self.protection += item["protection"]
+
+    def take_weapon(self) -> None:
+        if self.weapon:
+            self.power += self.weapon["power"]
+
+    def drink_potion(self) -> None:
+        if self.potion:
+            for name, value in self.potion["effect"].items():
+                self.__dict__[name] += value
+
+    def prepare_to_battle(self) -> None:
+        self.put_on_armor()
+        self.take_weapon()
+        self.drink_potion()
+
+    def attack(self, target: Hero) -> None:
+        target.hp -= self.power - target.protection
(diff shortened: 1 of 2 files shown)


app/hero.py
```diff
@@ -0,0 +1,87 @@
...
+            armor: list = None,
+            weapon: dict = None,
+            potion: dict = None
+    ) -> None:
+        self.name = name
+        self.power = power
+        self.hp = hp
+        self.armor = armor if armor else []
+        self.weapon = weapon if weapon else {}
+        self.potion = potion if potion else {}
```
MNDonut, line 19: you can just use "or" operator here

app/hero.py
```diff
@@ -0,0 +1,87 @@
...
+            potion=config.get("potion")
+        )
+
+    def put_on_armor(self) -> None:
+        if self.armor:
+            for item in self.armor:
+                self.protection += item["protection"]
+
+    def take_weapon(self) -> None:
+        if self.weapon:
```
MNDonut, line 39: weapon is required by the requirements

app/hero.py
```diff
@@ -0,0 +1,87 @@
...
+            hp=config.get("hp"),
+            armor=config.get("armour"),
+            weapon=config.get("weapon"),
+            potion=config.get("potion")
+        )
+
+    def put_on_armor(self) -> None:
+        if self.armor:
+            for item in self.armor:
+                self.protection += item["protection"]
```
MNDonut, line 36: you can remove if statement because armor is a list with data or empty list(as you set it in init), thus it'll loop through all armour or not because it's empty

MNDonut (changes requested):

tests/test_main.py
Oleksl888, line 1: no need to make changes to tests

app/main.py
```diff
@@ -1,214 +1,25 @@
-KNIGHTS = {
```
Oleksl888, line 1: where did the config go? I would put it in a separate file and then import from there

app/main.py
```diff
@@ -1,214 +1,25 @@
...
 
-    # apply armour
-    mordred["protection"] = 0
-    for a in mordred["armour"]:
-        mordred["protection"] += a["protection"]
+def battle(config: dict) -> dict:
 
-    # apply weapon
-    mordred["power"] += mordred["weapon"]["power"]
+    lancelot = Hero.create_from_config(config.get("lancelot"))
```
Oleksl888, line 6: you repeat same action 4 times - you need to use a cycle instead

app/hero.py
```diff
@@ -0,0 +1,85 @@
+from __future__ import annotations
+
+
+class Hero:
+    def __init__(
```
Oleksl888, line 5: it is overall not a good concept to have one superclass that does it all.
Your code should be more modular - create separate classes for equipment to offload some logic from here and make it more reusable

Oleksl888 (changes requested):

app/hero.py
```diff
@@ -0,0 +1,97 @@
+from __future__ import annotations
+
+from app.equipment.armor import Armor
```
Oleksl888, line 3: this creates a circular dependency because you import equipment here and then in equipment you import Hero. This is an antipattern. Refactor your code so your equipment doesn't have a dependency on Hero

Oleksl888 (changes requested):

PR #729, opened 2024-08-05
add knights

nmakivchuk:


app/main.py
```diff
@@ -1,214 +1,34 @@
...
-                "hp": +10,
-                "power": +5,
-            }
-        }
-    }
-}
+from typing import Dict
+from configs import KNIGHTS
+from knight import Knight
+from battle import battle
```
NazarSheremet, line 4:
```suggestion
from app.configs import KNIGHTS
from app.knight import Knight
from app. battle import battle
```

NazarSheremet (changes requested):
image
Incorrect import

app/main.py
nmakivchuk: Unfortunately your suggestion doesn't work too. the same error

NazarSheremet (approved):
Good job!

PR #736, opened 2024-08-12
'Solution'

RToister:


app/battle/battle_realisation.py
```diff
@@ -0,0 +1,66 @@
+from app.classes import Knight
+
+
+def battle(knights_config: dict) -> dict:
+
+    lancelot = Knight(
+        "Lancelot",
```
vasylhnatiuk, line 7: get name from dict as well

vasylhnatiuk: try to use loop instead of duplicating

vasylhnatiuk (changes requested):

PR #738, opened 2024-08-13
Solution11

jyjuk:


app/main.py
```diff
@@ -1,214 +1,73 @@
...
-                "hp": +10,
-                "power": +5,
-            }
-        }
-    }
+        "armour": [{"part": "breastplate", "protection": 25}],
+        "weapon": {"name": "Sword", "power": 45},
+        "potion": {"name": "Blessing", "effect": {"hp": 10, "power": 5}},
+    },
 }
```
LLkaia, line 42: Let's move config to another file like `config.py ` (and do not change data, please)

app/knight.py
```diff
@@ -0,0 +1,41 @@
...
+    ) -> None:
+        self.name = name
+        self.base_power = power
+        self.base_hp = hp
+        self.armour = armour
+        self.weapon = weapon
+        self.potion = potion
+        self.apply_armour()
+        self.apply_weapon()
+        self.apply_potion()
```
LLkaia, line 22: Let's move these setup functions to be running from another function like `setup()` and then this function will run these 3

app/knight.py
```diff
@@ -0,0 +1,41 @@
...
+        self.power = self.base_power + self.weapon["power"]
+
+    def apply_potion(self) -> None:
+        if self.potion:
+            effect = self.potion["effect"]
+            self.hp = self.base_hp + effect.get("hp", 0)
+            self.power += effect.get("power", 0)
+            self.protection += effect.get("protection", 0)
+        else:
+            self.hp = self.base_hp
```
LLkaia, line 37: Let's try another approach
```
        self.hp = self.base_hp
        if self.potion:
            effect = self.potion["effect"]
            self.hp += effect.get("hp", 0)
            self.power += effect.get("power", 0)
            self.protection += effect.get("protection", 0)
```

LLkaia (changes requested):

app/main.py
```diff
@@ -1,214 +1,35 @@
...
+def battle(knights_config: Dict[str, Dict]) -> Dict[str, int]:
+    knights = {
+        name: Knight(**config)
+        for name, config in knights_config.items()
+    }
+    results = {}
 
-    if mordred["hp"] <= 0:
-        mordred["hp"] = 0
+    # Lancelot vs Mordred
```
Dimosphen1, line 14: Remove these comments, code is self-explicit enough so there is no need for them

app/main.py
```diff
@@ -1,214 +1,35 @@
...
-    if arthur["potion"] is not None:
-        if "power" in arthur["potion"]["effect"]:
-            arthur["power"] += arthur["potion"]["effect"]["power"]
-
-        if "protection" in arthur["potion"]["effect"]:
-            arthur["protection"] += arthur["potion"]["effect"]["protection"]
-
-        if "hp" in arthur["potion"]["effect"]:
-            arthur["hp"] += arthur["potion"]["effect"]["hp"]
+from typing import Dict
```
Dimosphen1, line 1: Add one blank line between different groups of imports to follow PEP-8 here and in the other files

app/knight.py
```diff
@@ -0,0 +1,45 @@
...
+            effect = self.potion["effect"]
+            self.hp += effect.get("hp", 0)
+            self.power += effect.get("power", 0)
+            self.protection += effect.get("protection", 0)
+
+    def __str__(self) -> str:
+        return f"{self.name}: " \
+               f"HP={self.hp}, " \
+               f"Power={self.power}, " \
+               f"Protection={self.protection}"
```
Dimosphen1, line 45: It would be better to avoid using `\` and place string inside `()` for multi-line placement

app/battle.py
```diff
@@ -0,0 +1,18 @@
...
+        self.knight2 = knight2
+
+    def fight(self) -> Dict[str, int]:
+        self.knight1.hp -= max(0, self.knight2.power - self.knight1.protection)
+        self.knight2.hp -= max(0, self.knight1.power - self.knight2.protection)
+        self.knight1.hp = max(0, self.knight1.hp)
+        self.knight2.hp = max(0, self.knight2.hp)
+        return {
+            self.knight1.name: self.knight1.hp,
+            self.knight2.name: self.knight2.hp
```
Dimosphen1, line 17: You could return names in title case here and use the dict directly in the `update()` operation later

Dimosphen1 (changes requested):

PR #740, opened 2024-08-14
My solution

lavryniuk-dev:


app/main.py
```diff
@@ -1,214 +1,41 @@
...
-        if "hp" in red_knight["potion"]["effect"]:
-            red_knight["hp"] += red_knight["potion"]["effect"]["hp"]
+    knights = {}
+    for data in knights_config.values():
+        knight = Knight(data["name"],
+                        data["power"],
+                        data["hp"],
+                        parse_armour(data["armour"]),
+                        parse_weapon(data["weapon"]),
+                        parse_potion(data["potion"]))
```
LiudmylaKulinchenko, line 16: How about instead of passing reach parameter you'll do [unpacking](https://medium.com/@ashishkush1122/dictionary-unpacking-in-python-544f957e035a)
To make it work you'll need to move parsing armour, weapon and potion to Knight class and create them directly in init method

app/main.py
```diff
@@ -1,214 +1,41 @@
...
-        mordred["name"]: mordred["hp"],
-        red_knight["name"]: red_knight["hp"],
-    }
+def parse_weapon(weapon: dict) -> Weapon:
+    return Weapon(weapon["name"], weapon["power"])
 
 
-print(battle(KNIGHTS))
+def parse_potion(potion: dict) -> Potion:
+    return Potion(potion["name"], potion["effect"]) if potion else None
```
LiudmylaKulinchenko, line 41: You can use unpacking in these methods as well

LiudmylaKulinchenko (changes requested):

app/warrior/knight.py
```diff
@@ -52,3 +55,16 @@ def fight(self, opponent: Knight) -> Knight:
...
+        return [Armour(element["part"],
+                       element["protection"]) for element in armour]
+
+    @staticmethod
+    def parse_weapon(weapon: dict) -> Weapon:
+        return Weapon(**weapon)
+
+    @staticmethod
+    def parse_potion(potion: dict) -> Potion:
+        return Potion(**potion) if potion else None
```
LiudmylaKulinchenko, line 70: You can avoid creting these methods. Try to put it directly inside `__init__`

LiudmylaKulinchenko (approved):
Well done! 🌟

PR #746, opened 2024-08-20
Develop

AversArthur:


app/battle_results.py
```diff
@@ -0,0 +1,11 @@
+from app.knight import Knight
+
+
+def result(first: Knight, second: Knight) -> None:
```
vasylhnatiuk, line 4: it is redundant create file for only one function

app/battle_preparations.py
```diff
@@ -0,0 +1,11 @@
+from app.knight import Knight
+
+
+def battle_preparations(knight_dict: dict) -> list:
```
vasylhnatiuk, line 4: it is redundant create file for only one function

vasylhnatiuk: you can save this function in class Knight as class method

app/battle_results.py
vasylhnatiuk: you can save this function in class Knight as static method

app/knight.py
```diff
@@ -0,0 +1,27 @@
...
+        self.potion = potion
+        self.protection = protection
+
+    def get_ready_to_battle(self) -> None:
+        self.protection += sum(item.
+                               get("protection", 0)
+                               for item in self.armour)
+        self.power += self.weapon.get("power", 0)
+        if self.potion:
+            buff = self.potion.get("effect", {})
```
vasylhnatiuk, line 24: use loop and `settatr` instead of duplicating

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,214 +1,27 @@
...
-    # 2 Arthur vs Red Knight:
-    arthur["hp"] -= red_knight["power"] - arthur["protection"]
-    red_knight["hp"] -= arthur["power"] - red_knight["protection"]
+def battle(dict_of_knight: dict) -> dict:
+    fighter = Knight.battle_preparations(dict_of_knight)
 
-    # check if someone fell in battle
-    if arthur["hp"] <= 0:
-        arthur["hp"] = 0
+    lancelot = next(knight for knight in fighter if knight.name == "Lancelot")
```
vasylhnatiuk, line 8: don't use iterator just in battle_preparations write knights into dict and after get each knight by a key

app/knight.py
```diff
@@ -0,0 +1,48 @@
...
+
+        if self.potion:
+            buff = self.potion.get("effect", {})
+            for attr in ["power", "hp", "protection"]:
+                if attr in buff:
+                    current_value = getattr(self, attr)
+                    setattr(self, attr, current_value + buff[attr])
+
+    @classmethod
+    def battle_preparations(cls, knight_dict: dict) -> list:
```
vasylhnatiuk, line 32: return dict instead of list where a key is a knight name

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #753, opened 2024-08-30
Solution

Deadmarvald:


NazarSheremet (changes requested):
Good refactoring, but will be better to move class and input data in separate files

NazarSheremet (approved):
Good job!

PR #754, opened 2024-09-06
Solution

denys-badass:


NazarSheremet (approved):
Your solution is excellent, maybe a few overhead, no need to write doc string for every function

PR #762, opened 2024-09-17
Solution

vrudkovsky:


NazarSheremet (changes requested):
image
image

Function `battle` should get only one argument, as it was before refactoring

Losiev (approved):
Well done!👏

PR #767, opened 2024-09-25
created battle func and class Knight

Mxbely:


Atikiho (approved):
LGTM

app/knights/param.py
```diff
@@ -0,0 +1,32 @@
...
+
+        # apply armour
+        for armor in knight["armour"]:
+            self.protection += armor["protection"]
+
+        # apply weapon
+        self.power += knight["weapon"]["power"]
+
+        # apply potion if exist
+        if knight["potion"] is not None:
```
AnyoneClown, line 24: We can make this shorter `if knight["potion"]`

app/knights/param.py
```diff
@@ -0,0 +1,32 @@
...
+        self.name = name
+        self.hp = hp
+        self.power = power
+        self.protection = protection
+
+    def calculate_param(self, knight: dict) -> None:
+
+        # apply armour
+        for armor in knight["armour"]:
+            self.protection += armor["protection"]
```
AnyoneClown, line 18: We can use `sum()` here, refactor it

app/main.py
```diff
@@ -1,214 +1,50 @@
...
+        knights_config["mordred"]["power"]
+    )
+    mordred.calculate_param(knights_config["mordred"])
+
+    red_knight = Knight(
+        knights_config["red_knight"]["name"],
+        knights_config["red_knight"]["hp"],
+        knights_config["red_knight"]["power"]
+    )
+    red_knight.calculate_param(knights_config["red_knight"])
```
AnyoneClown, line 34: Try simplifying this code, for example, by assigning it all through a loop.


You may ask ‘Why do I need this’, but imagine that you want to change the names of your knights, you will have to change all these keys at the same time.

You can create for loop, where u will create knight instance and then u will store them inside of list/tuple(for example) and then you will call them on fight with list

AnyoneClown (changes requested):

Nattalli (approved):
👍🏻

PR #772, opened 2024-09-30
Solution

vladyslav-tmf:


KKvladys (approved):

app/knights.py
```diff
@@ -0,0 +1,73 @@
...
+        self.protection = sum(
+            armour_element["protection"]
+            for armour_element in self.armour
+        )
+
+    def apply_weapon(self) -> None:
+        self.total_power = self.base_power + self.weapon["power"]
+
+    def apply_potion(self) -> None:
+        if self.potion is not None:
```
AnyoneClown, line 56:
```suggestion
        if self.potion:
```

This style will be better

app/knights.py
```diff
@@ -0,0 +1,73 @@
...
+            for stat, value in self.potion["effect"].items():
+                if stat == "hp":
+                    self.hp += value
+                elif stat == "power":
+                    self.total_power += value
+                elif stat == "protection":
+                    self.protection += value
```
AnyoneClown, line 63: From python 3.10 we can implement switch/case functionality. U can implement it here.

[Link](https://www.freecodecamp.org/news/python-switch-statement-switch-case-example/)

AnyoneClown (changes requested):

app/knights.py
vladyslav-tmf: Done.

Nattalli (approved):

PR #773, opened 2024-10-01
implement py-knights-fighting

yuliiapyrih:


app/knight_info/knight.py
```diff
@@ -0,0 +1,58 @@
...
+    for knight_key, knight_value in knights_config.items():
+        current_knight = Knight(
+            knight_value["name"],
+            knight_value["hp"],
+            knight_value["power"]
+        )
+
+        current_knight.total_strength(
+            knight_value["armour"],
+            knight_value["weapon"],
```
LiudmylaKulinchenko, line 51: You can put armour and weapon creation directly inside Knight class init method
Then you could use unpacking (`**knight_data`)

app/knight_info/knight.py
```diff
@@ -0,0 +1,58 @@
...
+            weapon: dict,
+            potion: dict
+    ) -> Knight:
+        if potion is not None:
+            knight_potion = KnightPotion(potion["name"], potion["effect"])
+            self.hp += knight_potion.get_hp()
+            self.power += knight_potion.get_power()
+            self.protection += knight_potion.get_protection()
+
+        knight_weapon = KnightWeapon(weapon["name"], weapon["power"])
```
LiudmylaKulinchenko, line 27: Use unpacking where possible (using `**`)

app/knight_info/armour.py
```diff
@@ -0,0 +1,8 @@
+class KnightArmour:
+    @staticmethod
+    def total_armour(armours: list) -> int:
+        protection = 0
+        for armour in armours:
+            protection += armour["protection"]
+
+        return protection
```
LiudmylaKulinchenko, line 8: You can make it shorter using sum

app/knight_info/potion.py
```diff
@@ -0,0 +1,13 @@
+class KnightPotion:
```
LiudmylaKulinchenko, line 1: It's ok, but I would say not necessary to add Knight prefix to the classes, since it's a small project -- take it into consideration!

LiudmylaKulinchenko (changes requested):

app/knight_info/knight.py
yuliiapyrih: Done

app/knight_info/knight.py
yuliiapyrih: Done

app/knight_info/armour.py
yuliiapyrih: Done

app/knight_info/potion.py
yuliiapyrih: Done

app/knight_info/armour.py
```diff
@@ -0,0 +1,4 @@
+class Armour:
```
LiudmylaKulinchenko, line 1: Let's also create an init class here with storing armour name or "part"+ protection points!

LiudmylaKulinchenko (changes requested):
Small addition!

app/knight_info/armour.py
yuliiapyrih: Done

LiudmylaKulinchenko (approved):
Well done! 🌷🌼

PR #776, opened 2024-10-04
solution

illiachernets:


app/battle/battle_file.py
```diff
@@ -0,0 +1,14 @@
...
+    @staticmethod
+    def battle(first_knight: Knight, second_knight: Knight) -> None:
+        first_knight.hp -= second_knight.power - first_knight.protection
+        second_knight.hp -= first_knight.power - second_knight.protection
+
+        if first_knight.hp <= 0:
+            first_knight.hp = 0
+
+        if second_knight.hp <= 0:
+            second_knight.hp = 0
```
LLkaia, line 14: Use function `max()` in order to get rid of conditions

app/knight_info/potion.py
```diff
@@ -0,0 +1,13 @@
+class Potion:
+    def __init__(self, name: str, effect: dict) -> None:
+        self.name = name
+        self.effect = effect
+
+    def get_hp(self) -> int:
+        return self.effect["hp"] if "hp" in self.effect else 0
```
LLkaia, line 7: use get function with default value intead here, same for functions below

LLkaia: You even can avoid creation of this functions, use get instead

LLkaia (changes requested):

pavlopro (approved):
GJ!

PR #777, opened 2024-10-04
Solution

YaroslavBordovoy:


app/battle.py
```diff
@@ -0,0 +1,25 @@
+from typing import List
```
dxrrkwm, line 1: why not just use usual list annotation?

app/battle.py
```diff
@@ -0,0 +1,25 @@
...
+    @staticmethod
+    def start_battle(knights: List[Knight]) -> None:
+
+        # first fight
+        knights[0].hp -= knights[2].power - knights[0].protection
+        knights[2].hp -= knights[0].power - knights[2].protection
+
+        # second fight
+        knights[1].hp -= knights[3].power - knights[1].protection
+        knights[3].hp -= knights[1].power - knights[3].protection
```
dxrrkwm, line 17: i guess it's better to refer to knights by name, not index

app/knights/characters.py
dxrrkwm, line 1: not sure the name of the characters file is clear enough, i would use a different name

app/knights/knights.py
```diff
@@ -0,0 +1,19 @@
...
+
+    def __init__(
+            self,
+            name: str,
+            power: int,
+            hp: int,
+            armour: list = None,
+            weapon: dict = None,
+            potion: int = None
+    ) -> None:
```
dxrrkwm, line 11: you can use self and dict here

app/knights/knights.py
```diff
@@ -0,0 +1,19 @@
...
+
+        self.name = name
+        self.power = power
+        self.hp = hp
+        self.armour = armour
+        self.weapon = weapon
+        self.potion = potion
+        self.protection = 0
```
dxrrkwm, line 19: so here you will use keys of this dict, i guess it's a better practice in this case😇

dxrrkwm (changes requested):
everything is generally okay, add the proposed changes.  but i'm not sure if such a large-scale decomposition is necessary, maybe you can use fewer files, although i'm not sure if this is a best practice🥲

app/battle.py
YaroslavBordovoy: Fixed

app/battle.py
YaroslavBordovoy: I want to specify the list with the passed class instances explicitly. I think it's appropriate here. But I have corrected this annotation in other files

app/knights/characters.py
YaroslavBordovoy: Ok, I`ve changed it

app/knights/knights.py
YaroslavBordovoy: I agree, it means fewer lines of code

app/battle.py
```diff
@@ -8,13 +8,15 @@ class Battle:
...
-        knights[0].hp -= knights[2].power - knights[0].protection
-        knights[2].hp -= knights[0].power - knights[2].protection
+        lancelot.hp -= mordred.power - lancelot.protection
+        mordred.hp -= lancelot.power - mordred.protection
 
         # second fight
-        knights[1].hp -= knights[3].power - knights[1].protection
-        knights[3].hp -= knights[1].power - knights[3].protection
+        arthur.hp -= red_knight.power - arthur.protection
+        red_knight.hp -= arthur.power - red_knight.protection
```
dxrrkwm, line 19: you can also try to implement a method that will perform these actions to further increase readability

app/preparation/armour.py
```diff
@@ -1,11 +1,9 @@
...
-
 from app.knights.knights import Knight
 
 
 class Armour:
 
     @staticmethod
-    def get_armour(knights: List[Knight], armour: list) -> None:
+    def get_armour(knights: Knight, armour: list) -> None:
         for i in armour:
```
dxrrkwm, line 8: you can rename i to improve readability

dxrrkwm (approved):
looks nice :)

app/preparation/armour.py
```diff
@@ -0,0 +1,9 @@
+from app.knights.knights import Knight
+
+
+class Armour:
+
+    @staticmethod
+    def get_armour(knights: Knight, armour: list) -> None:
+        for i in armour:
+            knights.protection += i["protection"]
```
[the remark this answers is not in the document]
Nattalli: alsp specify list type

PR #778, opened 2024-10-04
Solution?

dxrrkwm:


Den-k0 (approved):

app/knights.py
```diff
@@ -0,0 +1,38 @@
...
+        self.power = params["power"]
+        self.hp = params["hp"]
+        self.armour = params["armour"]
+        self.weapon = params["weapon"]
+        self.potion = params["potion"]
+        self.protection = 0
+
+    def apply_armor(self) -> None:
+        for armor in self.armour:
+            self.protection += armor["protection"]
```
Nattalli, line 16: can you use list and list comprehension for that?

app/knights.py
```diff
@@ -12,8 +12,7 @@ def __init__(self, params: dict) -> None:
         self.protection = 0
 
     def apply_armor(self) -> None:
-        for armor in self.armour:
-            self.protection += armor["protection"]
+        self.protection = sum([armor["protection"] for armor in self.armour])
```
AnyoneClown, line 15: We can use simple generator for this. Even without list

`self.protection = sum(armor["protection"] for armor in self.armour)`

AnyoneClown (approved):
Good, check comment

app/knights.py
dxrrkwm: makes sense🫠

PR #781, opened 2024-10-07
Solution

sashasyrota:


app/Battle/battle.py
```diff
@@ -0,0 +1,30 @@
+from app.Battle_Preparations.apply_weapon import apply_weapon
+from app.Battle_Preparations.apply_armour import apply_armour
+from app.Battle_Preparations.apply_potion import apply_potion
+
+
+def battle(knights: dict) -> dict:
+    # apply weapon
```
Dimosphen1, line 7: Remove these comments, the code is self-explicit enough

app/Battle/__init__.py
Dimosphen1, line 1: Please, use lower case in directory names here and in the other places as well

app/Battle/battle.py
```diff
@@ -0,0 +1,30 @@
...
+        second_knight["hp"] -= (first_knight["power"]
+                                - second_knight["protection"])
+        if first_knight["hp"] <= 0:
+            first_knight["hp"] = 0
+        if second_knight["hp"] <= 0:
+            second_knight["hp"] = 0
+        result.update({
+            first_knight["name"]: first_knight["hp"],
+            second_knight["name"]: second_knight["hp"]
+        })
```
Dimosphen1, line 29: You could place part of the logic that happens in the loop to a separate function for a better readability

app/Battle_Preparations/apply_armour.py
Dimosphen1, line 1: There is no need to create separate file for each function, create a single file `utils.py` and place armour, potion and weapon functions there

app/main.py
```diff
@@ -85,130 +88,5 @@
...
-    return {
-        lancelot["name"]: lancelot["hp"],
-        arthur["name"]: arthur["hp"],
-        mordred["name"]: mordred["hp"],
-        red_knight["name"]: red_knight["hp"],
-    }
-
-
-print(battle(KNIGHTS))
+# battle
```
Dimosphen1, line 91: Remove redundant comments

app/main.py
```diff
@@ -1,4 +1,7 @@
-KNIGHTS = {
+from app.Battle.battle import battle
+
+
+knights = {
```
Dimosphen1, line 4: Place knights config to a separate file for better logic separation

app/main.py
```diff
@@ -85,130 +88,5 @@
...
-        lancelot["name"]: lancelot["hp"],
-        arthur["name"]: arthur["hp"],
-        mordred["name"]: mordred["hp"],
-        red_knight["name"]: red_knight["hp"],
-    }
-
-
-print(battle(KNIGHTS))
+# battle
+print(battle(knights))
```
Dimosphen1, line 92: It could be a good idea to use `if __name__ == "__main__"` here

Dimosphen1 (changes requested):

app/main.py
sashasyrota: зображення_2024-10-08_142647058
you meant place config like this?

pavlopro (approved):
GJ!

PR #797, opened 2024-10-17
Solution of knights fighting

serg121192:


app/main.py
```diff
@@ -1,214 +1,56 @@
...
+def knights_creation(config: dict) -> Knight:
+    character = Knight(config["name"], config["power"], config["hp"])
+    armours = []
+    if config["armour"] is not None:
+        for armour in config["armour"]:
+            armours.append(Armour(
+                armour.get("part"),
+                armour.get("protection")
+            ))
+    if config["potion"] is not None:
```
LLkaia, line 16: same here. you can create separate variable for `armor` and `potion`

LLkaia (changes requested):

PR #812, opened 2024-11-01
Ukraine, Russia, China, Iran

Greena1c:


app/main.py
```diff
@@ -1,214 +1,94 @@
+from typing import Dict, Any
+
 KNIGHTS = {
```
Alkaponees, line 3: Recommend remove this part from main.py and put this value to separate file

Alkaponees (changes requested):
Please check my  review

app/main.py
Alkaponees: I mean KNIGHTS

YuriiSmolii (approved):
Well done!

PR #826, opened 2024-11-18
Created all moduls and classes

shafran-andriy:


app/main.py
```diff
@@ -1,213 +1,78 @@
...
+        power=knights_config.get("lancelot").get("power"),
+        hp=knights_config.get("lancelot").get("hp"),
+        armour=knights_config.get("lancelot").get("armour"),
+        weapon=knights_config.get("lancelot").get("weapon"),
+        potion=knights_config.get("lancelot").get("potion"),
+        protection=sum(
+            [protect.get("protection")
+             for protect in knights_config.get("lancelot").get("armour")]
+        )
+    )
```
vsmutok, line 18:
```suggestion
    lancelot = Knight(**knights_config.get("lancelot"))
```

1. It's better to perform unpacking to reduce code duplication and improve its style.
2. Calculating protection is better moved to the Knight class, so it is done there and not repeated every time a new hero is created.

app/knights/battle.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+def fight(knight1: Knight, knight2: Knight) -> None:
+    knight1.hp -= knight2.power - knight1.protection
+    knight2.hp -= knight1.power - knight2.protection
+    # check if someone fell in battle
+    if knight1.hp <= 0:
+        knight1.hp = 0
+    if knight2.hp <= 0:
+        knight2.hp = 0
```
vsmutok, line 11:
```suggestion
        knight2.hp = 0
        
        return knight1, knight2
```

app/main.py
```diff
@@ -1,213 +1,78 @@
...
+            [protect.get("protection")
+             for protect in knights_config.get("red_knight").get("armour")]
+        )
+    )
+
+    for knight in Knight.instances:
+        # apply weapon
+        knight.apply_weapon()
+        # apply potion if exist
+        knight.apply_potion()
```
vsmutok, line 63: This piece of code is redundant; it is better to do it directly in the `__init__` method.

shafran-andriy: fixed

app/main.py
shafran-andriy: fixed

app/knights/battle.py
shafran-andriy: fixed

app/knights/battle_preparation.py
```diff
@@ -0,0 +1,41 @@
...
+        self.apply_weapon()
+        self.apply_potion()
+
+    def apply_weapon(self) -> None:
+        self.power += self.weapon.get("power")
+
+    def apply_potion(self) -> None:
+        # apply potion if exist
+        if self.potion is not None:
+            if "power" in self.potion.get("effect").keys():
```
vsmutok, line 33:
```suggestion
            if "power" in self.potion.get("effect"):
```
`keys()` - by default

vsmutok (changes requested):

app/knights/battle_preparation.py
```diff
@@ -0,0 +1,41 @@
...
+                 ) -> None:
+        self.name = name
+        self.power = power
+        self.hp = hp
+        self.armour = armour
+        self.weapon = weapon
+        self.potion = potion
+        self.protection = sum([protect.get("protection")
+                               for protect in self.armour])
+        Knight.instances.append(self)
```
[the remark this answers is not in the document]
shafran-andriy: It is not redundant here, because in this line we sum protection in the armour dist
and in the method apple_potion we add to protection from the "effect"

vasylhnatiuk (approved):

PR #849, opened 2024-12-05
'Solution'

Dogober:


taraskhom (approved):
Great job, looks like I checked my work instead of yours)

app/knight.py
```diff
@@ -0,0 +1,36 @@
+class Knight:
+
+    def __init__(self, knight: dict) -> None:
+        self.name = knight["name"]
+        self.power = knight["power"]
+        self.hp = knight["hp"]
+        self.armour = knight["armour"]
+        self.weapon = knight["weapon"]
+        self.potion = knight["potion"]
```
L1nk3rrr, line 9: it would be better to use external classes for armour, weapon, potion for better understanding what is it

app/knight.py
```diff
@@ -0,0 +1,36 @@
...
+            if self.potion["effect"].get("protection"):
+                self.protection += self.potion["effect"]["protection"]
+
+    def battle_preparation(self) -> None:
+        self.apply_armour()
+        self.apply_weapon()
+        self.apply_potion()
+
+    def fell_in_battle(self) -> None:
+        if self.hp <= 0:
```
L1nk3rrr, line 35: You can simple use `<` cause if our `hp` is already 0 no sense to set it again to 0

app/main.py
```diff
@@ -1,213 +1,47 @@
...
 
-        if "hp" in arthur["potion"]["effect"]:
-            arthur["hp"] += arthur["potion"]["effect"]["hp"]
+    lancelot = Knight(knights_config["lancelot"])
+    lancelot.battle_preparation()
 
-    # mordred
-    mordred = knightsConfig["mordred"]
+    arthur = Knight(knights_config["arthur"])
+    arthur.battle_preparation()
```
L1nk3rrr, line 12: you can create a list or some data that will contain all knights and in one loop prepare them for battle

app/main.py
```diff
@@ -1,213 +1,47 @@
...
     # Return battle results:
     return {
-        lancelot["name"]: lancelot["hp"],
-        arthur["name"]: arthur["hp"],
-        mordred["name"]: mordred["hp"],
-        red_knight["name"]: red_knight["hp"],
+        lancelot.name: lancelot.hp,
+        arthur.name: arthur.hp,
+        mordred.name: mordred.hp,
+        red_knight.name: red_knight.hp,
```
L1nk3rrr, line 44: if you will have in a start list of knights you can do it via dict comprehension!

app/knight.py
L1nk3rrr: also would be nice to create a module for knight and his ammunition, so all related files will be on the same package

L1nk3rrr (changes requested):

L1nk3rrr (approved):
Well done!

PR #850, opened 2024-12-06
Solution 

taraskhom:


Dogober (approved):
Wow. Looks like cool!!!!

L1nk3rrr (changes requested):
Well done! Check my comments for improvements of your code.
Also, in Python projects we have files in snake_case too, so `knightsConfig` must be `knights_config` and `Data` must be `data`

L1nk3rrr (changes requested):
knight and knight related classes better to move for related package, so in you app would be main file and package with data and package with knight related stuff

PR #851, opened 2024-12-07
Solution

ruslanyemets:


taraskhom (approved):
Amazingly done, I am delighted

app/main.py
```diff
@@ -1,214 +1,47 @@
...
 
-    # apply weapon
-    lancelot["power"] += lancelot["weapon"]["power"]
+        knight.apply_weapon(weapon)
 
-    # apply potion if exist
-    if lancelot["potion"] is not None:
-        if "power" in lancelot["potion"]["effect"]:
-            lancelot["power"] += lancelot["potion"]["effect"]["power"]
+        if knights_config.get(name, {}).get("potion", {}) is not None:
```
L1nk3rrr, line 33: `get("potion", {})` do you need a default dict here?

app/main.py
```diff
@@ -1,214 +1,47 @@
...
+            potion = Potion(
+                knights_config.get(name, {}).get("potion", {})["name"],
+                knights_config.get(name, {}).get("potion", {})["effect"]
```
L1nk3rrr, line 36: you can use a variable instead of code duplication.
Also you take "name" and "effect" without get, what if it would be not at your dict?

app/arena/fight.py
```diff
@@ -0,0 +1,16 @@
...
+        knight1.hp -= knight2.power - knight1.protection
+        knight2.hp -= knight1.power - knight2.protection
+        knight1.hp = Fight.check_hp(knight1.hp)
+        knight2.hp = Fight.check_hp(knight2.hp)
+
+    @staticmethod
+    def check_hp(hp: int) -> int:
+        if hp < 0:
+            return 0
+        return hp
```
L1nk3rrr, line 16: hmm, nice!
I think it would be better to implement this method at knight, for checking hp

app/knights/knight.py
```diff
@@ -0,0 +1,23 @@
+from app.knights.armour import Armour
+from app.knights.weapon import Weapon
+from app.knights.potion import Potion
+
+
+class Knight:
+    protection = 0
```
L1nk3rrr, line 7: maybe it must be an instance attribute not a class?

app/main.py
```diff
@@ -1,214 +1,47 @@
...
+        name: Knight(knight["name"], knight["power"], knight["hp"])
+        for name, knight in knights_config.items()
     }
-}
 
+    for name, knight in participants.items():
 
-def battle(knightsConfig):
-    # BATTLE PREPARATIONS:
+        if knights_config.get(name, {}).get("armour", []) is not None:
```
L1nk3rrr, line 17: `.get("armour", []) ` do you need here default list?

L1nk3rrr (changes requested):
Great work!
I left some comments for improvements (about fight logic - you can try, but if not -> it's good, because it's my proposal not a task requirement, you already did a great job!)
Check for same items about `.get...` in your code also.

L1nk3rrr (approved):
Well done!

PR #854, opened 2024-12-08
My solution

DanilTrukhanov:
Created classes for knights, potion, armour, and weapon objects and made them separate modules. Made knights_config a separate module.
For class, knight added methods to get a weapon, put an armor on and drink potion (if available). Made a class method to create Knight instances from a knights_config file.


ihorhalyskiy (changes requested):
fix flake8 error

ihorhalyskiy (approved):

PR #886, opened 2025-01-21
Solution

sosunovych-andrii:


app/knight.py
```diff
@@ -0,0 +1,31 @@
...
+        self.armour = knight_data.get("armour", [])
+        self.weapon = knight_data.get("weapon")
+        self.potion = knight_data.get("potion", None)
+
+        self.protection = self.apply_armour()
+        self.apply_weapon()
+        self.apply_potion()
+
+    def apply_armour(self) -> int:
+        return sum([item.get("protection", 0) for item in self.armour])
```
vasylhnatiuk, line 18:
```suggestion
        return sum(item.get("protection", 0) for item in self.armour)
```

app/knight.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def apply_weapon(self) -> None:
+        self.power += self.weapon.get("power", 0)
+
+    def apply_potion(self) -> None:
+        if self.potion is not None:
+            for buff, value in self.potion.get("effect").items():
+                if buff == "power":
```
vasylhnatiuk, line 26: try to use settatr instead of many if-else

vasylhnatiuk (changes requested):

PR #918, opened 2025-02-20
Solution 2

oukiz:
I changed: OOP, import from `knights_date`, methods for preparation, battle, and tournament, processing knights through methods, battle via `knights_battle`, result as winner's name, automated tournament.
I really hope it's enough <3


vasylhnatiuk (changes requested):
зображення
fix tests

PR #919, opened 2025-02-21
Completed task knights-fighting

Yurkiv313:


app/main.py
```diff
@@ -1,213 +1,67 @@
...
+    for armour_data in knight_data["armour"]:
+        fighter.add_armour(Armour(**armour_data))
+
+    if knight_data["potion"]:
+        fighter.add_potion(Potion(**knight_data["potion"]))
+
+    return fighter
+
+
+def battle(knights_config: dict) -> dict:
```
BronkstoneBro, line 25: I didn't complete this additional task, but it seems to me that this function is extremely difficult to describe in terms of the fact that it might need to be modified someday, and it has hardcoding. But as I understand it, this is a requirement of the task.

BronkstoneBro (approved):

app/main.py
Yurkiv313: yes, that name was in the task before

PR #925, opened 2025-02-24
Solution 

iSevenpwnz:


Kart0l (approved):
If the protection is greater than the attack power, we will get a positive value, which is incorrect.
And maybe it's better to use classes.
But the general code works correctly, good work.

PR #930, opened 2025-02-26
Solution

HappyDen08:


app/battle.py
```diff
@@ -0,0 +1,38 @@
...
+def battle(knights: dict) -> dict:
+    stats = {}
+    for key, knight in knights.items():
+        stats[key] = calculate_stats(knight)
+
+    stats["lancelot"]["hp"] -= (
+        stats["mordred"]["power"] - stats["lancelot"]["protection"]
+    )
+    stats["mordred"]["hp"] -= (
+        stats["lancelot"]["power"] - stats["mordred"]["protection"]
```
Boyok1337, line 13: better approach to use `.get()` to avoid `KeyError`. Here you have hardcoded value, so `KeyError` not happen, but for future make it a habit for you

app/battle.py
```diff
@@ -0,0 +1,38 @@
...
+        stats["lancelot"]["hp"] = 0
+    if stats["mordred"]["hp"] < 0:
+        stats["mordred"]["hp"] = 0
+
+    stats["arthur"]["hp"] -= (
+        stats["red_knight"]["power"] - stats["arthur"]["protection"]
+    )
+    stats["red_knight"]["hp"] -= (
+        stats["arthur"]["power"] - stats["red_knight"]["protection"]
+    )
```
Boyok1337, line 26: same here

app/battle.py
```diff
@@ -0,0 +1,38 @@
...
+        stats["mordred"]["power"] - stats["lancelot"]["protection"]
+    )
+    stats["mordred"]["hp"] -= (
+        stats["lancelot"]["power"] - stats["mordred"]["protection"]
+    )
+
+    if stats["lancelot"]["hp"] < 0:
+        stats["lancelot"]["hp"] = 0
+    if stats["mordred"]["hp"] < 0:
+        stats["mordred"]["hp"] = 0
```
Boyok1337, line 19: same here

app/battle.py
Boyok1337: Try smth like `a.get("key", {}).get("key", 0)`

app/battle.py
```diff
@@ -0,0 +1,38 @@
...
+        stats["red_knight"]["power"] - stats["arthur"]["protection"]
+    )
+    stats["red_knight"]["hp"] -= (
+        stats["arthur"]["power"] - stats["red_knight"]["protection"]
+    )
+
+    if stats["arthur"]["hp"] < 0:
+        stats["arthur"]["hp"] = 0
+    if stats["red_knight"]["hp"] < 0:
+        stats["red_knight"]["hp"] = 0
```
Boyok1337, line 31: check everywhere

app/litsar.py
```diff
@@ -0,0 +1,20 @@
+def calculate_stats(knight: dict) -> dict:
+    effective_hp = knight["hp"]
+    effective_power = knight["power"] + knight["weapon"]["power"]
+    effective_protection = sum(
+        item.get("protection", 0) for item in knight.get("armour", [])
+    )
+
+    potion = knight.get("potion")
```
Boyok1337, line 8: good use of `.get()`

app/litsar.py
```diff
@@ -0,0 +1,20 @@
...
+    if potion is not None:
+        effect = potion.get("effect", {})
+        effective_hp += effect.get("hp", 0)
+        effective_power += effect.get("power", 0)
+        effective_protection += effect.get("protection", 0)
```
Boyok1337, line 13: Good one!

app/settings.py
```diff
@@ -0,0 +1,66 @@
+KNIGHTS = {
```
Boyok1337, line 1: also, you can divide to each knight and add to dict, then Its gonna be easy to read and edit if you have hardcoded values

Boyok1337 (changes requested):

Boyok1337 (approved):
Good job!

PR #934, opened 2025-02-28
Solution

cemen2009:


app/AppPackage/Characters/BaseCharacter.py
```diff
@@ -0,0 +1,45 @@
...
+from app.AppPackage.Items.Weapon import Weapon
+
+
+class BaseCharacter:
+    def __init__(self, character_data: dict) -> None:
+        self.name = character_data["name"]
+        self.power = character_data["power"]
+        self.hp = character_data["hp"]
+        self.armour = Armour(character_data["armour"])
+        self.weapon = Weapon(character_data["weapon"])
```
Boyok1337, line 11: A good approach with classes.
BaseCharacter is more suitable for a class that will act as an interface and from which you can implement all the Characters. In your case, you should change the name to Character.

app/AppPackage/Characters/BaseCharacter.py
```diff
@@ -0,0 +1,45 @@
...
+        if character_data.get("potion") is not None:
+            self.apply_potion(character_data["potion"]["effect"])
+
+    def apply_potion(self, effect_data: dict) -> None:
+        if effect_data.get("power") is not None:
+            self.power += effect_data["power"]
+        if effect_data.get("hp") is not None:
+            self.hp += effect_data["hp"]
+        if effect_data.get("protection") is not None:
+            self.armour.protection += effect_data["protection"]
```
Boyok1337, line 22: Use `.get()`everywhere, cause it can trigger KeyError and program stop

app/AppPackage/Items/Armour.py
```diff
@@ -0,0 +1,24 @@
...
+        self.helmet_protection: int = Armour.find_armour("helmet", armour_data)
+        self.breastplate_protection: int = Armour.find_armour(
+            "breastplate", armour_data
+        )
+        self.boots_protection: int = Armour.find_armour(
+            "boots", armour_data
+        )
+        self.protection = (self.helmet_protection
+                           + self.breastplate_protection
+                           + self.boots_protection)
```
Boyok1337, line 15: each item from new line

app/main.py
```diff
@@ -1,3 +1,6 @@
+from app.AppPackage.Characters.BaseCharacter import BaseCharacter
+
+
 KNIGHTS = {
```
Boyok1337, line 4: move to other file like knight_config.py

Boyok1337 (changes requested):

app/AppPackage/Items/Armour.py
cemen2009: otherwise flake8 will trigger because line is more than 80 characters

app/AppPackage/Characters/BaseCharacter.py
cemen2009: i check ther using get to make sure that this key is exist.
i don't use directly `.get()` because it could raise "TypeError: unsupported operand type(s) for +: 'int' and 'NoneType'"

Boyok1337: Ok, using .get(key, 0) will prevent the TypeError because if the key is missing, it defaults to 0, ensuring the arithmetic operation works without issues.

Boyok1337 (approved):
Good one, but google how to avoid TypeError errors using .`get()`

PR #944, opened 2025-03-08
Solution

Oneyura:


Kart0l (approved):
Considering the topic to which this task relates, it would be better to move the calculation of total power and protection into separate methods.
But in global sense the code works.

PR #950, opened 2025-03-12
Solution

LLevkovych:


Boyok1337 (changes requested):
1. Use .get() to avoid KeyError when accessing optional fields.
2. Separate the class and function
3. User construction `if "__name__" == "__main__"` to run file

PR #962, opened 2025-03-20
Solution

chiper9:


app/models/potion.py
```diff
@@ -0,0 +1,16 @@
+class Potion:
+    def __init__(self, name: str, effect: bool) -> None:
+        self.name = name
+        self.effect = effect
+
+    def apply(self, knight: dict) -> None:
+        if self.effect:
```
MNDonut, line 7: it's better to do reversed condition so that you'll simplify nested statements

MNDonut (approved):

PR #1051, opened 2025-06-20
Solution

DrGregoryHouse4676:


app/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+    "name": "Katia",
+    "age": 26,
+    "wallet": wallet_3,
+}
+
+john = {
+    "name": "John",
+    "age": 29,
+    "wallet": wallet_3,
+}
```
MNDonut, line 16: what is it?

MNDonut (changes requested):

app/__init__.py
DrGregoryHouse4676: I just don't remember if there should be anything there, too much time has passed.

PR #1059, opened 2025-06-25
solution

zqpiq:
My solution could be improved — it could be implemented using classes, and some parts could be refactored into separate functions. However, even though the final result was identical to the original code, the tests didn't accept my solution.

### DIFF
--- a/app/preparation.py
+++ b/app/preparation.py
@@ -15,7 +15,7 @@ def preparation(dictionary_knights: dict) -> list[dict]:
 
             if "protection" in knight["potion"]["effect"]:
                 knight["protection"] += (
-                    knight)["potion"]["effect"]["protection"]
+                    knight["potion"]["effect"]["protection"])
 
             if "hp" in knight["potion"]["effect"]:
                 knight["hp"] += knight["potion"]["effect"]["hp"]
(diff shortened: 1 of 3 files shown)


app/preparation.py
```diff
@@ -0,0 +1,22 @@
...
+        for defend in knight["armour"]:
+            knight["protection"] += defend["protection"]
+        knight["power"] += knight["weapon"]["power"]
+        if knight["potion"] is not None:
+            if "power" in knight["potion"]["effect"]:
+                knight["power"] += knight["potion"]["effect"]["power"]
+
+            if "protection" in knight["potion"]["effect"]:
+                knight["protection"] += (
+                    knight)["potion"]["effect"]["protection"]
```
viannik, line 18: The parentheses are misplaced, could cause a syntax error

app/preparation.py
```diff
@@ -0,0 +1,22 @@
+def preparation(dictionary_knights: dict) -> list[dict]:
```
viannik, line 1: The preparation function modifies the original knight dictionaries in place, which may have unintended side effects if called multiple times or reused.

viannik (approved):
The PR meets most requirements for modularity and clarity, but fixing the mutation bug and potentially adopting classes would make it exemplary. 

p.s. Tests were run locally and passed successfully both before and after fixing the syntax bug.

PR #1064, opened 2025-06-29
added solution with OOP approach

Orixara:


app/models/knight.py
```diff
@@ -0,0 +1,50 @@
...
+    def prepare_for_battle(self) -> None:
+        if self.weapon:
+            self.power += self.weapon.power
+        self.calculate_protection()
+        self.apply_potion_effects()
+
+    def take_damage(self, damage: int) -> None:
+        self.hp -= damage
+        if self.hp < 0:
+            self.hp = 0
```
dimashevchukk, line 47: can use max(self.hp, 0)

app/battle/battle.py
dimashevchukk, line 1: if you got logic related to class knight it is better to keep it inside as a methods

app/data/knights_data.py
dimashevchukk, line 1: looks too complicated, like u r about to create a whole new For Honnor game, but that's on u

dimashevchukk (approved):

PR #1070, opened 2025-07-01
Implemented task

GrishaSeniv:


app/models/armour.py
```diff
@@ -0,0 +1,23 @@
...
+        armours = []
+
+        for armour in data:
+            armours.append(Armour.from_dict(armour))
+
+        return armours
+
+    @classmethod
+    def from_dict(cls, data: dict) -> Armour:
+        return cls(part=data.get("part"), protection=data.get("protection"))
```
MNDonut, line 23: how do you access part if this class doesn't have this attribute?

app/models/armour.py
```diff
@@ -0,0 +1,23 @@
+from __future__ import annotations
+
+
+class Armour:
+    def __init__(self, part: str, protection: int) -> None:
+        self.par = part
+        self.protection = protection
```
MNDonut, line 7: here you have `par` attribute

app/models/armour.py
MNDonut, line 23: but here you set `part` attribute

app/main.py
```diff
@@ -86,129 +88,27 @@
...
-        mordred["name"]: mordred["hp"],
-        red_knight["name"]: red_knight["hp"],
-    }
+def build_result(knights: list[Knight]) -> dict:
+    return {knight.name: knight.hp for knight in knights}
 
 
-print(battle(KNIGHTS))
+if __name__ == "__main__":
+    print(battle(KNIGHTS))
```
MNDonut, line 114: we don't need it here

app/models/effect.py
```diff
@@ -0,0 +1,22 @@
+from __future__ import annotations
+
+
+class Effect:
+    def __init__(self,
+                 power: int = None,
+                 hp: int = None,
+                 protection: int = None) -> None:
```
MNDonut, line 8: fix annotation. you set `None` values to `int` types

app/models/knight.py
```diff
@@ -0,0 +1,76 @@
...
+
+class Knight:
+    def __init__(self,
+                 name: str,
+                 power: int,
+                 hp: int,
+                 weapon: Weapon,
+                 armour: list[Armour] = None,
+                 potion: Potion = None,
+                 protection: int = 0) -> None:
```
MNDonut, line 16: fix annotation

app/models/knight.py
```diff
@@ -0,0 +1,76 @@
...
+        self.name = name
+        self.power = power
+        self.hp = hp
+        self.weapon = weapon
+        self.protection = protection
+        self.armour = armour if armour is not None else []
```
MNDonut, line 22: just `if not armour` or it's better `if armour`

app/models/knight.py
```diff
@@ -0,0 +1,76 @@
...
+
+        if effect.hp:
+            self.hp += effect.hp
+
+    def battle(self, other: Knight) -> None:
+        self.hp -= other.power - self.protection
+        other.hp -= self.power - other.protection
+
+        Knight.check_if_fell(self)
+        Knight.check_if_fell(other)
```
MNDonut, line 57: that is a bit weird that you use static method to not have a context of a Knight object, but pass this object. Isn't it better to make it a usual method?

MNDonut (changes requested):

PR #1081, opened 2025-07-12
Solution

SensibleN00B:


Vskesha (approved):
Great code separation into classes. The code is clear and understandable. Great work.

SensibleN00B:
@mateacademy-ai-mentor but if knights potion is None, I create Potion instance with zero parametrs. In this case it cant influence to knights characteristics

PR #1132, opened 2025-09-17
py-knights-fighting

EvilAtomic:


vasylhnatiuk (changes requested):
delete all comments and docs in rus!! all should be in English

PR #1140, opened 2025-09-21
Solution

delirium95:


app/config/config.py
```diff
@@ -0,0 +1,86 @@
+KNIGHTS = {
```
Vorongor: Good idea - take the initial data into a separate file.

PR #1331, opened 2026-03-22
Implemented THE BEST :) solution for py-knights-fighting

Neyrony:


Neyrony:
i don't agree with this case
   if knight1.stats.hp < 0:
        knight1.stats.hp = 0
    if knight2.stats.hp < 0:
        knight2.stats.hp = 0
Imagine the situation when they killed each other with one hit and they both less then 0 hp, i think my var better

PR #1351, opened 2026-04-02
Solution

alisahoys:


Sacchar20 (approved):

andriihrechko (approved):

app/battle.py
```diff
@@ -0,0 +1,8 @@
+from typing import Dict
+from app.knight import Knight
+
+
+def fight(knight1: Knight, knight2: Knight) -> Dict[str, int]:
```
VadimPusiak788, line 5: This functionality should be in the file main.py because the fight function is related to the battle function, and they should be in the same context.

VadimPusiak788 (changes requested):
Please

PR #1352, opened 2026-04-03
Solution v1.0

andriihrechko:


alisahoys (approved):
Clear separation of items and knight. Easy to read. Tie case not handled

PR #1359, opened 2026-04-07
Solution

BlackSun333:


app/main.py
```diff
@@ -1,214 +1,70 @@
...
+
+        if self.potion:
+            effect = self.potion.get("effect", {})
+            self.hp += effect.get("hp", 0)
+            self.power += effect.get("power", 0)
+            self.protection += effect.get("protection", 0)
+
+    def receive_damage(self, opponent_power: int) -> None:
+        damage = opponent_power - self.protection
+        if damage > 0:
```
VadimPusiak788, line 37: You can use max() here to avoid the second if block:

`self.hp = max(0, self.hp - damage)`

app/main.py
```diff
@@ -1,214 +1,70 @@
...
-}
 
+    lancelot = knights["lancelot"]
+    mordred = knights["mordred"]
+    arthur = knights["arthur"]
+    red_knight = knights["red_knight"]
 
-def battle(knightsConfig):
-    # BATTLE PREPARATIONS:
+    lancelot_final_power = lancelot.power
```
VadimPusiak788, line 55: Battle logic between two knights is repeated — extract into a fight(knight1, knight2) helper function

VadimPusiak788 (changes requested):

PR #1363, opened 2026-04-10
Refactored main.py, added modules: combat.py, config.py, entities.py

zwirok-git:


app/config.py
```diff
@@ -0,0 +1,92 @@
...
+            "effect": {
+                "hp": +10,
+                "power": +5,
+            },
+        },
+    },
+}
+
+
+DUELS = (
```
alisahoys, line 89: good config-driven design

app/battle/combat.py
```diff
@@ -0,0 +1,9 @@
+from app.battle.entities import Knight
+
+
+def run_duel(first_knight: Knight, second_knight: Knight) -> None:
```
alisahoys, line 4: Encapsulation keeps combat rules centralized

alisahoys (approved):
good structure, clean separation of files

PR #1439, opened 2026-06-21
feat: implement battle func() and battle/knights packages

neeeksa:


app/main.py
vsmutok: Move the dictionary with data to a separate module.

vsmutok (changes requested):

